PR #5, opened 2017-07-09
Typos and testing the tool chain

tsundberg:
When testing the tool chain, I saw a few typos.

I don't think pom.xml should not be capitalized when the file that it refers to isn't capitalized.

A Gherkin example was not properly indented.

### DIFF
--- a/content/tutorial/getting-started-with-cucumber-jvm-in-five-minutes.md
+++ b/content/tutorial/getting-started-with-cucumber-jvm-in-five-minutes.md
@@ -63,7 +63,7 @@ Let's take a quick look at the project structure:
 
 Now we'll look a bit closer at the following parts of the structure and tidy up a bit:
 
-* In the skeleton directory, you'll see a _POM.XML_ file 
+* In the skeleton directory, you'll see a _pom.xml_ file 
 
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
I think this looks exactly right!

olleolleolle:
(Meta thought: Is a typo fix a "bug" in a repo with docs?)

tsundberg:
From my perspective, a typo is a bug in docs.

The documentation should be correct and if it isn't then there is a bug that should be fixed.

aslakhellesoy:
Thanks @tsundberg. The current content will be scrapped in favour of all the contents on the `sw-reorg` branch. This branch contains a dump of docs from a variety of sources.

When we're ready to accept improvements to the contents we'll announce it - we're currently only focussing on collecting all the old docs in one place.

tsundberg:
I'm looking forward to the day when the content is in such a shape that we can add improrovements through pull requests.

Cheers!

plaindocs:
Hi @tsrosenberg you're welcome to make PRs against the `sw-reorg` branch now, just bare in mind that it is very in much in beta at the moment, and looks pretty different to what was here before.

I'm fixing up some last minute stuff so we can merge it into master.

PR #43, opened 2017-10-10
Fix number formatting in css

mlvandijk:
This seems to do the trick.
See preview: https://deploy-preview-43--cucumber.netlify.com/cucumber/browser-automation/

### DIFF
--- a/themes/cucumber-hugo/static/css/cucumber.css
+++ b/themes/cucumber-hugo/static/css/cucumber.css
@@ -3515,19 +3515,14 @@ input[type="submit"].button {
   -webkit-box-align: center;
       -ms-flex-align: center;
           align-items: center;
-  background-color: whitesmoke;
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
The preview you showed looked very regular.

Good change!

plaindocs:
Thanks for fixing @mlvandijk ! 

`prism.js` is usually generated, so perhaps it was a rogue plugin for that.

http://prismjs.com/plugins/line-highlight/ kinda thing.

mlvandijk:
@plaindocs If it's generated then it might be generated again? in which case this fix will not hold... Will figure that out should it become relevant :)

plaindocs:
Exactly, that is why I left this breadcrumb. ;-)

Prism is the JS syntax highlighting library, and afaik you select the languages+plugins you want and download a JS blob.

PR #47, opened 2017-10-16
Update support of JVM languages

mlvandijk:
Most JVM languages have been removed from cucumber-jvm to their own repo.
(see: https://github.com/cucumber/cucumber-jvm)
Update the documentation in line with this change.


olleolleolle (approved):
Great and clear update.

PR #55, opened 2017-10-29
Update jvm group id, versions, options and add JUnit annotations

mlvandijk:
Update group id from info.cukes to io.cucumber
Update to latest jvm version and snapshot version
Update command line options
Add supported JUnit annationations (related to https://github.com/cucumber/cucumber-jvm/issues/1268)

### DIFF
--- a/content/implementations/jvm.md
+++ b/content/implementations/jvm.md
@@ -189,7 +189,7 @@ java cucumber.api.cli.Main --version
 The [JUnit Runner](#junit-runner) and [Android Runner](#android-runner) can also pick
 up configuration options defined via the `@CucumberOptions` annotation.
 
-**Note:** Cucumber only supports @ClassRule},@BeforeClass and @AfterClass JUnit annotations.
+**Note:** Cucumber only supports @ClassRule,@BeforeClass and @AfterClass JUnit annotations.
 
 For example, if you want to tell Cucumber to use the two formatter plugins `pretty` and `html`, you can specify it like this:
 
--- a/content/implementations/jvm.md
+++ b/content/implementations/jvm.md
@@ -30,17 +30,18 @@ Other JVM languages have been moved to their own repository:
 
 ## Installation
 
-Cucumber-JVM consists of several modules (JARs) that you can download from the [public Maven repo](http://repo1.maven.org/maven2/io/cucumber/).
-There is no "setup" program for Cucumber-JVM---just JAR files.
+Cucumber-JVM consists of several modules (JARs). We recommend using a build tool, like [Maven](https://maven.apache.org/) or [Gradle](https://gradle.org/),
+to manage the dependencies in your project.
 
-Determining which JARs to add to your project's `CLASSPATH` depends on the programming language you are using. If you
-are using Java, you may want to add one of the JARs for [dependency injection](/implementations/jvm/java-di/) as well.
+Determining which dependencies to add to your project, depends on the programming language you are using. If you
+are using Java, you may also want to add one of the JARs for [dependency injection](/implementations/jvm/java-di/).
+In addition, you may want to use a test framework, like JUnit or TestNg.
 
 ### Snapshot releases
 
 If you want to take advantage of functionality that has been committed to the git `master` branch, but hasn't been released to the public maven repo yet, you can use `SNAPSHOT` builds from the [sonatype snapshot repo](https://oss.sonatype.org/content/repositories/snapshots/io/cucumber/).
 
-Just add the following to your `pom.xml`:
+If you are using Maven, just add the following to your `pom.xml`:
 
 ```xml
 <repository>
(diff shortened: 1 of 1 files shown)


content/implementations/jvm.md
```diff
@@ -189,6 +189,8 @@ java cucumber.api.cli.Main --version
 The [JUnit Runner](#junit-runner) and [Android Runner](#android-runner) can also pick
 up configuration options defined via the `@CucumberOptions` annotation.
 
+**Note:** Cucumber only supports @ClassRule},@BeforeClass and @AfterClass JUnit annotations.
```
olleolleolle: Is the { significant here?

mlvandijk: Nope, copy-paste mistake. Removed it, thanks!

content/implementations/jvm.md
```diff
@@ -58,9 +58,9 @@ For example:
 
 ```xml
 <dependency>
-    <groupId>info.cukes</groupId>
+    <groupId>io.cucumber</groupId>
     <artifactId>cucumber-java</artifactId>
-    <version>{{ site.versions.cucumber_jvm }}-SNAPSHOT</version>
+    <version>2.1.1-SNAPSHOT</version>
```
mpkorstanje, line 64: Shouldn't this use  site.versions.cucumber_jvm ?

content/implementations/jvm.md
```diff
@@ -102,7 +102,7 @@ To use the JUnit runner you need to add the following dependencies:
 <dependency>
     <groupId>info.cukes</groupId>
     <artifactId>cucumber-junit</artifactId>
-    <version>{{ site.versions.cucumber_jvm }}</version>
+    <version>2.1.0</version>
```
mpkorstanje: GroupId should be io.cucumber

content/implementations/jvm.md
```diff
@@ -387,9 +395,9 @@ Definition.
 
 ```xml
 <dependency>
-    <groupId>info.cukes</groupId>
+    <groupId>io.cucumber</groupId>
     <artifactId>cucumber-groovy</artifactId>
-    <version>{{ site.versions.cucumber_jvm }}</version>
+    <version>${version}</version>
```
mpkorstanje: May actually need two versions. One for the versions that are no longer part of core and one for the versions that are.

content/implementations/jvm.md
```diff
@@ -278,9 +286,9 @@ Definitions, you need:
 
 ```xml
 <dependency>
-    <groupId>info.cukes</groupId>
-    <artifactId>cucumber-java8</artifactId>
-    <version>{{ site.versions.cucumber_jvm }}</version>
+    <groupId>io.cucumber</groupId>
+    <artifactId>cucumber-java</artifactId>
```
mpkorstanje: This should be java8

content/implementations/jvm.md
```diff
@@ -189,6 +189,8 @@ java cucumber.api.cli.Main --version
 The [JUnit Runner](#junit-runner) and [Android Runner](#android-runner) can also pick
 up configuration options defined via the `@CucumberOptions` annotation.
 
+**Note:** Cucumber only supports @ClassRule,@BeforeClass and @AfterClass JUnit annotations.
```
mpkorstanje: Single back ticks might help to clarify that the annotations are code.

content/implementations/jvm.md
```diff
@@ -19,26 +19,26 @@ Please see the [General Reference](/gherkin/gherkin-reference/) for functionalit
...
+- [JRuby](https://github.com/cucumber/cucumber-jvm-jruby)
+- [Jython](https://github.com/cucumber/cucumber-jvm-jython)
+- [Rhino](https://github.com/cucumber/cucumber-jvm-rhino)
+- [Scala](https://github.com/cucumber/cucumber-jvm-scala)
 
 
 ## Installation
 
-Cucumber-JVM consists of several modules (JARs) that you can download from the [public Maven repo](http://repo1.maven.org/maven2/info/cukes/).
+Cucumber-JVM consists of several modules (JARs) that you can download from the [public Maven repo](http://repo1.maven.org/maven2/io/cucumber/).
```
mpkorstanje: Perhaps out of scope for this PR but I think we should recommend maven/gradle or any other dependency management tool here. Don't want people downloading jars manually.

mpkorstanje: Definitely need Maven introduced here. Explanations further on rely on it.

content/implementations/jvm.md
mlvandijk: Or we should (re)move the dependencies for the jvm languages that were removed from cucumber-jvm?

content/implementations/jvm.md
mlvandijk: Working on it ;)

content/implementations/jvm.md
mlvandijk: Should that show the current version? That doesn't seem to work atm (nor on the current site). Fyi, for the `SNAPSHOT` version, it's using an actual version number in the example, rather than a variable...

mpkorstanje: You might have a look at this: https://gohugo.io/variables/ could be in there I think

Issue #17, opened 2017-09-03
Cucumber JVM docs issues with versions while using maven


naveenb29:
I followed the docs and ran into an issue similar to this - 

https://stackoverflow.com/questions/25291480/cucumber-exception-java-lang-classnotfoundexception-cucumber-io-resourceloader

I needed to make 2 fixes add the `cucumber-core` to my dependencies and also increment the `cucumber-junit` version to `cucumber-java's` versions (2.0.X)

I would like to send a pull for this if the community thinks this is a valid concern for beginners.

plaindocs:
Hi @naveenb29 

Thanks for looking into that. We're *heavily* refactoring the docs in https://github.com/cucumber/docs.cucumber.io/pull/3 so make sure you branch from `sw-reorg` before making the changes.  :-)

And for sure! Examples that don't work are no fun for anyone

mlvandijk:
@naveenb29 Branch sw-reorg has just been merged to master; you can create a PR to master

mlvandijk:
Fixed by PR #76

PR #76, opened 2017-11-11
Add note to use same version numbers

mlvandijk:
Should solve #17

### DIFF
--- a/themes/cucumber-hugo/layouts/shortcodes/tip.html
+++ b/themes/cucumber-hugo/layouts/shortcodes/tip.html
@@ -1,4 +1,4 @@
-<article class="message is-primare">
+<article class="message is-primary">
   <div class="message-header">
     <p>{{ .Get 0 }}</p>
   </div>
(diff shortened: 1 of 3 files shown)


aslakhellesoy (changes requested):
Can you convert this into a `{{% tip %}}` box instead?

Or better, make a red `{{% note %}}` box. Copy the `tip.html` shortcode to `note.html` and style it with red. See bulma docs for what css class to use.

mlvandijk:
@aslakhellesoy I've added a {{% note %}} shortcode and tagged the notes as such. 
Red was a bit much imho, so I opted to style in black ('is-dark') for now. Found the bulma info: https://bulma.io/documentation/components/message :)

themes/cucumber-hugo/layouts/shortcodes/tip.html
```diff
@@ -1,4 +1,4 @@
-<article class="message is-primary">
+<article class="message is-primare">
```
olleolleolle: Is this a typo or the real class name?

olleolleolle:
Looks good now!

PR #77, opened 2017-11-14
Merge files to good-practices, update info and add polyglot

mlvandijk:
Merge 'step-organization' and 'anti-patterns' into 'good-practices'.
Update/edit text.
Add polyglot examples to page.


content/good-practices.md
```diff
@@ -1,3 +1,188 @@
 ---
+source: https://github.com/cucumber/cucumber/wiki/Step-Organization/
 title: Good Practices
```
mpkorstanje: Perhaps out of scope but I think it should be best practices. There can be multiple good practices but only one is best.

PR #83, opened 2017-12-08
Dont link out at the top of the file like that

plaindocs:
- people will think it is a TOC

### DIFF
--- a/content/browser-automation.md
+++ b/content/browser-automation.md
@@ -5,13 +5,7 @@ title: Browser Automation (Old)
 polyglot: true
 ---
 
-Cucumber is not a Browser Automation tool, but it works well with Browser
-Automation tools, such as:
-
-- [Selenium Webdriver](http://docs.seleniumhq.org/projects/webdriver/)
-- [Capybara](http://teamcapybara.github.io/capybara/)
-- [Watir](http://watir.com)
-- [Serenity](http://serenity-bdd.info)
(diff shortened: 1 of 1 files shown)


content/browser-automation.md
```diff
@@ -1,17 +1,11 @@
 ---
 menu: reference
 renderer: Cucumber::Website::Reference
```
mlvandijk, line 3: Remove 'renderer:' as I don't think it's used anymore?

content/browser-automation.md
```diff
@@ -89,6 +83,8 @@ Then("^the page title should start with "([^"]*)"$") do
 end
 ```
 
+More information on [Selenium Webdriver](http://docs.seleniumhq.org/projects/webdriver/).
+
 # Watir Webdriver
```
mlvandijk, line 88: Watir is Ruby specific so show only for Ruby page?

content/browser-automation.md
```diff
@@ -229,6 +227,14 @@ public class GoogleHomePage extends PageObject {
 public class SearchResultsPage extends PageObject {}
 ```
 
+More information on [Serenity](http://serenity-bdd.info).
+
+# Capybara
```
mlvandijk, line 232: Capybara is Ruby specific so show only for Ruby page

content/browser-automation.md
plaindocs: I don't think so if we're listing tools like this - best to show people what there is. We should however make it more clear which are language specific

PR #86, opened 2017-12-10
Testable architecture

mlvandijk:
First version of Testable Architecture, of what I think is intended here, based on several blogs and podcasts.
Sources are listed under "More information:" but could be moved to blog-posts (and recommended media?).

Please review to see if I got the right meaning and didn't over simplify in my summary.

### DIFF
--- a/content/testable-architecture.md
+++ b/content/testable-architecture.md
@@ -5,7 +5,7 @@ description: Designing for testability
 
 Automated tests and executable specifications can only work with software that has a testable architecture.
 A testable architecture enables fast and reliable tests that are easy to write, execute and maintain.
-Fast tests enable the developers to run them and obtain fast feedback on what they are building, without loosing focus or flow.
+Fast tests enable the developers to run them and obtain fast feedback on what they are building, without losing focus or flow.
 
 When designing for testability, make sure that your products and services are composed of loosely-coupled,
 well-encapsulated components or modules; following the principles of
@@ -32,7 +32,7 @@ an adapter for the queue, and an adapter for the web service. There is also a po
 a web server which plugs into this port and displays a UI in a browser. All the IO tends to happen outside of these ports.
 
 This architectural patterns is called the **ports and adapters pattern** (or [hexagonal architecture](http://alistair.cockburn.us/Hexagonal+architecture)).
-This pattern enables you to hook your scenarios and unit tests at a lower level, while he 'contract tests' give you the confidence to do that.
+This pattern enables you to hook your scenarios and unit tests at a lower level, while the 'contract tests' give you the confidence to do that.
 
 You’ll want to run *some* tests that go through the whole depth of your stack, to get complete confidence.
 Diagnosing where the problem is in a full-stack, end-to-end test is really hard, because it can be anywhere.
--- a/content/testable-architecture.md
+++ b/content/testable-architecture.md
@@ -3,4 +3,31 @@ title: Testable Architecture
 description: Designing for testability
 ---
 
-Testable Architecture
+Automated tests and executable specifications can only work with software that has a testable architecture.
+
+A testable architecture enables fast and reliable tests that are easy to write, execute and maintain.
+
+Having fast tests enable developers to stay focused. The tests will tell them that they are
+building the right thing. In order to get into the flow, the tests need to give them really fast feedback.
+
+When designing for testability, make sure that your products and services are composed of loosely-coupled,
+well-encapsulated components or modules; follow the principles of
+[SOLID design](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design))).
+
+The following are essential for a testable architecture:
+- decouple your domain logic from your infrastructure so that you can test at different levels, with maximum confidence and minimum cost.
+- ports and adapters ([hexagonal architecture](http://alistair.cockburn.us/Hexagonal+architecture)), contract testing and the [test pyramid](https://martinfowler.com/bliki/TestPyramid.html)
+
+More information:
+https://www.nngroup.com/articles/response-times-3-important-limits/
+https://cucumber.io/blog/2017/06/29/fast-tests
+https://joecolantonio.com/testtalks/180-ten-years-cucumber-bdd-aslak-hellesoy/
+https://www.joecolantonio.com/2017/11/30/3-ways-ui-based-cucumber-bdd-can-go-bad/
+https://github.com/subsecondtdd/todo-subsecond
+https://www.youtube.com/watch?v=Fk4rCn4YLLU&feature=youtu.be
+https://speakerdeck.com/npryce/having-our-cake-and-eating-it-1
+http://alistair.cockburn.us/Hexagonal+architecture
+https://xp2017.sched.com/event/91jK/testable-architecture
+https://skillsmatter.com/skillscasts/9971-testable-software-architecture-with-aslak-hellesoy
+https://speakerdeck.com/aslakhellesoy/testable-architecture-at-devlin-2017
+https://skillsmatter.com/skillscasts/9971-testable-software-architecture-with-aslak-hellesoy
--- a/content/testable-architecture.md
+++ b/content/testable-architecture.md
@@ -5,38 +5,41 @@ description: Designing for testability
 
 Automated tests and executable specifications can only work with software that has a testable architecture.
 A testable architecture enables fast and reliable tests that are easy to write, execute and maintain.
-Fast tests enable the developers to run them and obtain fast feedback on what they are building, without losing focus or flow.
+Fast tests enable the developers to run them frequently to obtain fast feedback on what they are building, without losing focus or flow.
 
 When designing for testability, make sure that your products and services are composed of loosely-coupled,
-well-encapsulated components or modules; following the principles of
-[SOLID design](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design)).
+well-encapsulated components or modules.
 
 Decoupling your business logic from your infrastructure (the often slow and brittle components),
 makes it possible to test at different levels, with maximum confidence and minimum cost.
 
-* Ensure that you can run your tests underneath your UI. Do not depend solely on UI tests. They are slow, expensive, and hard to fix.
-* Running tests against a database makes tests slow. And to make the tests behave consistently each time we run them,
-we need to make sure the database is in the expected state for each test, so the tests don't interfere with one another.
+* Ensure that you can run your tests without going through the user interface (UI). They are slow, brittle, expensive, and hard to fix. Do not depend solely on UI tests.
+* Running tests against a database makes tests slow. We need to make sure the database is in the expected state before each test in order to make the tests behave consistently.
+This will ensure that tests don't interfere with each other and can be executed in any order.
 
-We want our tests to use some kind of in-memory stub implementation, instead of an actual database.
+We want most of our tests to use some kind of in-memory stub implementation, instead of an actual database.
 From the perspective of the domain logic, the behavior looks exactly the same.
 To make sure that the stub works in the same way as the real thing, we need to have confidence, which we can obtain using [contract tests](https://martinfowler.com/bliki/IntegrationContractTest.html).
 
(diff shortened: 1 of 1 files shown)


content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
 description: Designing for testability
 ---
 
-Testable Architecture
\ No newline at end of file
+Automated tests and executable specifications can only work with software that has a testable architecture.
+A testable architecture enables fast and reliable tests that are easy to write, execute and maintain.
+Fast tests enable the developers to run them and obtain fast feedback on what they are building, without loosing focus or flow.
```
olleolleolle: loosing => losing

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
...
+Decoupling your business logic from your infrastructure (the often slow and brittle components),
+makes it possible to test at different levels, with maximum confidence and minimum cost.
+
+* Ensure that you can run your tests underneath your UI. Do not depend solely on UI tests. They are slow, expensive, and hard to fix.
+* Running tests against a database makes tests slow. And to make the tests behave consistently each time we run them,
+we need to make sure the database is in the expected state for each test, so the tests don't interfere with one another.
+
+We want our tests to use some kind of in-memory stub implementation, instead of an actual database.
+From the perspective of the domain logic, the behavior looks exactly the same.
+To make sure that the stub works in the same way as the real thing, we need to have confidence, which we can obtain using ‘contract tests’.
```
olleolleolle: Is the ornamentation around contract tests exactly what you want, here?

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
...
+and storing and retrieving data. We can run our tests using both the stub and the real implementation. This gives us confidence that the stub behaves like the real thing.
+
+The business logic is at the core of the application, completely decoupled from external devices and services. It doesn’t know anything about databases, message queues,
+and web services because we’ve isolated them through these ports. Often, these ports are connected to adapters that interact with the real implementation. There will be an adapter for the database,
+an adapter for the queue, and an adapter for the web service. There is also a port that defines how we can interact with our system, such as
+a web server which plugs into this port and displays a UI in a browser. All the IO tends to happen outside of these ports.
+
+This architectural patterns is called the **ports and adapters pattern** (or [hexagonal architecture](http://alistair.cockburn.us/Hexagonal+architecture)).
+This pattern enables you to hook your scenarios and unit tests at a lower level, while he 'contract tests' give you the confidence to do that.
```
olleolleolle: he => the

olleolleolle (commented):
I commented on tiny details of this article, which I think is a good article. It introduces the area.

content/testable-architecture.md
mlvandijk: Not sure what you mean here. Do you mean the ' ' around 'contract tests'? Copied that from the https://sdjournal.org/bdd-like-jazz/ blogpost.

olleolleolle: I mean the single quotes, yes.

Is "contract tests" not official naming, then perhaps single quotes are fine, but if we wish to make it stand out, then bold type could perhaps be helpful.

olleolleolle: See, I'm all weasel words and equivocating. Sorry about that.

mlvandijk: No problem. Thank you for proof reading all the time! It's hard to see the details when you've written and read the same doc several times. I *really, really* appreciate you taking the time to review; it makes the end result that much better!

mlvandijk: Not sure if 'contract tests' is the 'official' name (see, doing it again with the ' '). There seems to be a lot of misunderstanding about different types of tests all the time. I might just rewrite a little bit and avoid the term 'contract tests' (unless someone can provide a link/definition).

mlvandijk:
@olleolleolle I've added a link to some more information about contract testing, which I hope helps a little for people unfamiliar with the term.

olleolleolle (approved):
Thanks for introducing this area.

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
 description: Designing for testability
 ---
 
-Testable Architecture
\ No newline at end of file
+Automated tests and executable specifications can only work with software that has a testable architecture.
+A testable architecture enables fast and reliable tests that are easy to write, execute and maintain.
+Fast tests enable the developers to run them and obtain fast feedback on what they are building, without losing focus or flow.
```
tsundberg: ... enable the developers to run them and obtain ...

could use an additional word 

... enable the developers to run them continuously and obtain ...

... enable the developers to run them often and obtain ...

... enable the developers to run them frequent and obtain ...


I'm not sure if the extra word should be 'continuously', 'often' or 'frequent'

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
...
+
+When designing for testability, make sure that your products and services are composed of loosely-coupled,
+well-encapsulated components or modules; following the principles of
+[SOLID design](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design)).
```
tsundberg: I would end this paragraph 

... services are composed of loosely-coupled, well-encapsulated components or modules.

I wouldn't include the SOLID principles here. They could be added as a resource at the end instead.

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
...
+
+Decoupling your business logic from your infrastructure (the often slow and brittle components),
+makes it possible to test at different levels, with maximum confidence and minimum cost.
+
+* Ensure that you can run your tests underneath your UI. Do not depend solely on UI tests. They are slow, expensive, and hard to fix.
```
tsundberg: I would replace 'underneath' with 'below'.
I would also not use the abbrevation UI. I would introduce it as
... below your user interface, UI.

I would add brittle to the list of properties for a UI test. 

They are slow, brittle, expensive, and hard to fix.

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
...
+* Running tests against a database makes tests slow. And to make the tests behave consistently each time we run them,
+we need to make sure the database is in the expected state for each test, so the tests don't interfere with one another.
+
```
tsundberg: I would turn around this sentence: 

"And to make the tests behave consistently each time we run them, we need to make sure the database is in the expected state for each test, so the tests don't interfere with one another."

Something like this:

We need to make sure the database is in the expected state for each test in order to make the tests behave consistently. This will ensure that the tests don't interfere with each other and can be executed in any order.

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
...
+We want our tests to use some kind of in-memory stub implementation, instead of an actual database.
```
tsundberg: I would rephrase "We want our tests to use some kind of in-memory stub implementation, instead of an actual database." as

"We want most of our tests to use some kind of in-memory stub implementation, instead of an actual database."

Making it a bit softer.

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
...
+From the perspective of the domain logic, the behavior looks exactly the same.
```
tsundberg: I would replace 'looks' with 'is'

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
...
+To make sure that the stub works in the same way as the real thing, we need to have confidence, which we can obtain using [contract tests](https://martinfowler.com/bliki/IntegrationContractTest.html).
+
+We can think of our database and our stub as two things that can be plugged into the same socket, or **port**.
+The application containing the business logic doesn't need to know it's talking to a stub rather than the actual database; it's simply talking to this port
```
tsundberg: I would add 'if' in "doesn't need to know it's talking" 

so it becomes

"doesn't need to know if it is talking"


I would also consider replacing 'rather than' with 'or'

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
...
+and storing and retrieving data. We can run our tests using both the stub and the real implementation. This gives us confidence that the stub behaves like the real thing.
+
+The business logic is at the core of the application, completely decoupled from external devices and services. It doesn’t know anything about databases, message queues,
+and web services because we’ve isolated them through these ports. Often, these ports are connected to adapters that interact with the real implementation. There will be an adapter for the database,
```
tsundberg: I would change the 'and' in the enumeration of techniques to 'or' as in

"It doesn’t know anything about databases, message queues, and web services because ..."

would become

"It doesn’t know anything about databases, message queues, or web services because ..."

I would also use the complete words 'we have' instead of 'we’ve'

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
...
+an adapter for the queue, and an adapter for the web service. There is also a port that defines how we can interact with our system, such as
+a web server which plugs into this port and displays a UI in a browser. All the IO tends to happen outside of these ports.
```
tsundberg: I'm not sure I understand what you mean in this sentense:

"There is also a port that defines how we can interact with our system, such as a web server which plugs into this port and displays a UI in a browser. All the IO tends to happen outside of these ports."

I think it is the beginning that confuses me, 'There is also a port'

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
...
+
+This architectural patterns is called the **ports and adapters pattern** (or [hexagonal architecture](http://alistair.cockburn.us/Hexagonal+architecture)).
+This pattern enables you to hook your scenarios and unit tests at a lower level, while the **contract tests** give you the confidence to do that.
```
tsundberg: I would remove the word 'the' before ports and adapter. Can't motivate more than I think it reads better.

I would replace 'This pattern' with 'It' and thus avoid the word This twice in two sentences next to each other.

I would use the word 'connect' instead of 'hook'

content/testable-architecture.md
```diff
@@ -3,4 +3,62 @@ title: Testable Architecture
...
+
+You’ll want to run *some* tests that go through the whole depth of your stack, to get complete confidence.
+Diagnosing where the problem is in a full-stack, end-to-end test is really hard, because it can be anywhere.
+These tests are brittle because one change can break all your tests. They are also slow,
+as there is IO involved when going through a a browser, web service or database.
```
tsundberg: I would consider replacing "can break all your tests" with "can break all of them"

tsundberg (approved):
I think this can be merged.

content/testable-architecture.md
```diff
@@ -3,4 +3,67 @@ title: Testable Architecture
...
+
+* Ensure that you can run your tests without going through the user interface (UI). They are slow, brittle, expensive, and hard to fix. Do not depend solely on UI tests.
+* Running tests against a database makes tests slow. We need to make sure the database is in the expected state before each test in order to make the tests behave consistently.
+This will ensure that tests don't interfere with each other and can be executed in any order.
+
+We want most of our tests to use some kind of in-memory stub implementation, instead of an actual database.
+From the perspective of the domain logic, the behavior looks exactly the same.
+To make sure that the stub works in the same way as the real thing, we need to have confidence, which we can obtain using [contract tests](https://martinfowler.com/bliki/IntegrationContractTest.html).
+
+We can think of our database and our stub as two things that can be plugged into the same socket, or **port**.
```
aslakhellesoy, line 24: I think the word "socket" can easily be misunderstood as "TCP socket", so maybe clarify that it is a "socket", not a socket :-)

aslakhellesoy (commented):
Excellent job @mlvandijk! You nailed it!

content/testable-architecture.md
mlvandijk: Thanks @aslakhellesoy ! Yeah, the "socket" here is used as a metaphor; I'll just remove it to avoid confusion.

PR #99, opened 2018-01-05
Moved the documentation on how to run from a test framework before how to run from a build tool

tsundberg:
Running from a build tool is possible when using Java. But the easiest integration is to use a test framework. I think it is better to show the most important way of running first. My assumption is that this will reduce the number of frustrated users that will need help in the future.

This doesn't change anything for the JavaScript or Ruby documentation. They don't use test framework integration's to execute Cucumber.

### DIFF
--- a/content/configuration.md
+++ b/content/configuration.md
@@ -55,6 +55,79 @@ Cucumber does not work when installed globally because cucumber needs to be requ
 
 {{% /block %}}
 
+{{% block "java" %}}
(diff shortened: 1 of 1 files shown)


mlvandijk (approved):
Thanks @tsundberg for the improvement!

olleolleolle:
@tsundberg This is caring about the reader! THANK YOU!

PR #110, opened 2018-02-04
Move links

mlvandijk:
* Remove the lists of links at the end of `testable-architecture` and `bdd-tutorial`.
* Add relevant blogs to `blog-posts`
* Link from "Who does what" to internal "Example mapping" page, instead of directly to the blog. (Link to blog is also provided at the end of that page)


olleolleolle (approved):
Great improvement. 

Much less outbound links from the BDD article means less overload for the reader.

mlvandijk:
Thanks for the quick review @olleolleolle !

PR #146, opened 2018-04-20
More landing page

plaindocs:
Is this an improvement? 

Needs some CSS tweaks

I think we should add the menu to the front page.


plaindocs:
@mattwynne I fixed the build issue. Suggesting *something* like this https://deploy-preview-146--cucumber.netlify.com/

tsundberg:
Updated the example so it doesn't say 'blah' anymore

mlvandijk:
@plaindocs Yes, great improvement. We'd need a link to API page though (or remove for now?). Same for Guides.
@tsundberg Thanks for example update.

PR #150, opened 2018-04-21
[DO NOT MERGE]: Polyglot **example** for tools

mlvandijk:
**NOTE:** This PR is for discussion; please do not merge until we compare with other option (TODO).

As discussed on Open Source Day today, we're not happy with the current pages that provide  information for different programming language domains, for example: https://docs.cucumber.io/tools/

This PR provides *an* example of how we could deal with this, using the polyglot option. Another option would be to have a kind of landing page for tools, with links to language-specific pages for Java tools / JavaScript tools / Ruby tools.

Each option has upsides (`+`) and downsides (`-`).

Background info:
- TOC (left menu) is generated based on headers
- Headers inside language blocks do not end up in TOC
=> Build will fail if there are headers inside a language block

**1. Polyglot:**
`+` Uses language tabs, so when navigating from other pages, the user will automatically end up on tab for selected language
`+` No more irrelevant info (i.e. belonging to other language)
`-` Cannot use headers within language blocks; therefore titles inside language blocks will not end up in the TOC on the left (e.g. **Maven**, **Rake**)
`-` As these language specific titles are no longer headers, we can no longer link directly to them.

Preview link: https://deploy-preview-150--cucumber.netlify.com/tools/

**2. Language specific pages** (TODO: I'll create a separate PR, so we can compare the two)
`+` Can contain language specific headers in TOC
`+` Can link directly to language specific tools (e.g. **Maven**, **Rake**)
`+` No more irrelevant info (i.e. belonging to other language) / "have exactly the right content for the given platform" - @mattwynne 
`-` "end up on a left-nav for that [] tools page, but with no context around it to say where I just came from, or how to get back there. So I feel a bit lost." - @mattwynne 
`-` User will have to specifically click a link to go to language specific page with desired info. 

Atm I'm not sure what would be easier to maintain...

### DIFF
--- a/content/cucumber/continuous-integration.md
+++ b/content/cucumber/continuous-integration.md
@@ -22,7 +22,7 @@ We recommend executing Cucumber from one of these on a CI server.
 
 # Rake
 
-If your CI server is [using Rake](/tools/#rake), just make sure
+If your CI server is [using Rake](/tools/), just make sure
 that Rake executes your Cucumber task.
 
 # Ant JUnit output
--- a/content/cucumber/state.md
+++ b/content/cucumber/state.md
@@ -89,7 +89,7 @@ World(MyHelper, MyOtherHelpers)
 
 This will `extend` each new `World` object with those modules.
 
-If you use [Ruby on Rails](/tools/#ruby-on-rails), there is already a `World` set up for you, so you will get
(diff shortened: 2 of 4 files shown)


mlvandijk:
Closing this PR as #151 is more in line with what we discussed and what @aslakhellesoy is doing for instance in the Installation page.

PR #151, opened 2018-04-22
Language pages for tools

mlvandijk:
Another option for dealing with pages that have language specific information.

As described in #150 - Let's discuss which option we prefer.


mattwynne:
Preview link: https://deploy-preview-151--cucumber.netlify.com/tools/

mattwynne:
What I like about this is that you're able to have exactly the right content for the given platform, and in the case of something like tools, it's obviously going to be very different, and you're not constrained by worrying about the header structure.

What I don't like is that you get a bit lost. When you go from the general `tools` page (which is linked from the main drop-down nav) to the `ruby-tools` page, say, you end up on a left-nav for that Ruby tools page, but with no context around it to say where I just came from, or how to get back there. So I feel a bit lost.

mlvandijk:
Thanks for the feedback @mattwynne!
Let's see if we can provide that feedback (i.e. a link back to "generic" tools? / buttons for the other languages?)
Overall, I personally prefer this option to #150, but would like feedback from @diabolo @plaindocs @aslakhellesoy and possibly others?

Note: overview of pros and cons per option in #150 :)

mlvandijk:
Or we do like @aslakhellesoy did for the [Installation page](https://docs.cucumber.io/installation/). Very nice! :)

mlvandijk:
I'll rebase this off master to resolve conflicts and see if I can add some buttons like @aslakhellesoy did on the Installation page. And then check throughout the docs to do the same solution everywhere...

aslakhellesoy:
I'll merge this. We'll soon have section pages like [this](https://doc.akka.io/docs/akka/current/guide/index.html) (automatically generated from files in a directory) - see #157.

mlvandijk:
Thanks @aslakhellesoy :)

PR #156, opened 2018-04-24
Adds a docker compose file to work with container

haroon-sheikh:


README.md
```diff
@@ -30,6 +30,10 @@ Build and serve the website:
 
     ./docker_shell.sh
 
+or you can just use docker-compose to bring up the container:
```
olleolleolle: Tiny thing: avoid words like “just”, “simply”, “obviously”.

olleolleolle (commented):
This is a thoughtful assistance to people who want to change the documentation and see the effects it has.

README.md
haroon-sheikh: Removed.

aslakhellesoy:
What does this provide beyond the existing `docker_shell.sh` shell script? The shell script lets you specify an arbitrary command, such as `docker_shell.sh bash` for an interactive session, or `docker_shell.sh make htmlproofer` to check all the links. How would you do that with `docker-compose`?

aslakhellesoy:
@haroon-sheikh what are your thoughts on this?

haroon-sheikh:
@aslakhellesoy You can do `docker-compose exec web bash` for interactive session, or `docker-compose exec web make htmlproofer`. It's just a better way of working with your docker containers. If we're all happy with using the docker_shell, I don't mind us sticking that way.

xtrasimplicity:
I'd be leaning the same way as @aslakhellesoy here. I love using `docker-compose` for apps with numerous containers, but I don't see any benefit over the shell script in this instance. :)

haroon-sheikh:
closing as per above.

PR #172, opened 2018-05-18
Update configuration and reference for cucumber-jvm 3.0.0

mpkorstanje:
I've added a really short description on how to configure the type registry in cucumber-jvm.

In doing so I noticed that the configuration page it self only described cucumber ruby. And that there is no configuration info at all for cucumber js. I've just added an empty page for that.

Some links still need to be fixed and while checking those links I noticed that there is yet another configuration section in the api.md


mlvandijk:
Once done/merged, please redirect from wiki to new page(s):
* https://github.com/cucumber/cucumber/wiki/Environment-Variables
* https://github.com/cucumber/cucumber/wiki/cucumber.yml

mlvandijk:
Should #175 be merged here? Is there anything else needed (apart from fixing broken links)?

mpkorstanje:
There is a structure problem with these docs.  The configuration is in two places and it's not the right configuration.

mpkorstanje:
Raining now so I've got some time.

mpkorstanje:
@mlvandijk How do I call the wizard and make him  show me a preview page?

mpkorstanje:
Nevermind found the wizard (in the emerald city).

content/cucumber/api.md
```diff
@@ -763,13 +747,88 @@ Cucumber does not work when installed globally because cucumber needs to be requ
 
 You can also run features using a [build tool](/tools/general#build-tools) or an [IDE](/tools/general#ides).
 
-# Configuration
 
-Cucumber provides several configuration options.
+## JUnit
```
mlvandijk, line 722: JUnit section probably should be moved to Java tools page. We could link to it here. (This might be outside the scope of the PR; if you want to do it, please do. Otherwise, create an issue and we'll do it later)

content/cucumber/configuration.md
```diff
@@ -1,13 +1,75 @@
 ---
 title: Cucumber Configuration
-subtitle: cucumber.yml, environment variables
+subtitle: cucumber.yml, type registry
```
mlvandijk: For Ruby, environment variables are still relevant, so maybe we should leave them in.
Overall I think the config page should be split into separate pages per language, like the Tools pages. I'll explain why in following comments.

content/cucumber/configuration.md
```diff
@@ -1,13 +1,75 @@
...
 polyglot: true
 ---
 
+
+# Type Registry Configuration
+
+{{% block "java" %}}
+The type registry is used to configure parameter and data table types. It can be configured by placing an implementation
```
mlvandijk: "parameter and data table types" is a little confusing to me; do you mean "parameters" or "parameter types". Please specify as otherwise it makes the reader wonder.

content/cucumber/configuration.md
```diff
@@ -1,13 +1,75 @@
...
+
+# Type Registry Configuration
+
+{{% block "java" %}}
+The type registry is used to configure parameter and data table types. It can be configured by placing an implementation 
+of `cucumber.api.TypeRegistryConfigurer` on the glue path.
+
+
+```java
+public class TypeRegistryConfiguration implements TypeRegistryConfigurer {
```
mlvandijk: Really appreciate having an example here. I feel it could use a little explanation though. Let me know if you want some help with that ;)

content/cucumber/configuration.md
```diff
@@ -1,13 +1,75 @@
...
+                }
+            })
+        );
+    }
+}
+```
+{{% /block %}}
+
+{{% block "ruby" %}}
+ TODO:
```
mlvandijk: Personally, I don't think we should have "TODO" anywhere in the documentation. If you cannot add this information yourself, please try to find someone with Ruby knowledge who can do it (or help you do it), or (if that fails) create an issue marked "Ruby"/"Help wanted"

content/cucumber/configuration.md
```diff
@@ -1,13 +1,75 @@
...
+{{% /block %}}
+
+{{% block "javascript" %}}
+ TODO:
```
mlvandijk: Again, I'd prefer no "TODO" on master.
If we split the Config page into page per language, Javascript section with only "TODO" would not be needed; we could simply open an issue that it remains to be added.

content/cucumber/configuration.md
```diff
@@ -22,17 +84,15 @@ want to execute with this profile.
...
-{{% text "ruby" %}}
 Cucumber-Rails creates a `cucumber.yml` file in the project config directory containing a number of predefined profiles,
 one of which is the default profile. When Cucumber is run from the command line, it is usually necessary to provide both
 the directory name containing the root directory of the tree containing feature files and the directory name containing
 references to the necessary library files. In a typical project, `cucumber --require features features/some/path` will suffice.
 Repetitious usages can be added to user-defined profiles contained in the project's `cucumber.yml` file.
-{{% /text %}}
 
-## Executing Profiles
+<b>Profile can be executed</b>
```
mlvandijk: You might not be aware, but we agreed earlier to not have headers that look like headers but aren't technically headers. We also do not allow headers inside language blocks (as you probably found out). The fact that we have headers here that are / seem language specific is a hint that the page should be split into language specific pages.

content/cucumber/configuration.md
```diff
@@ -55,7 +115,7 @@ output.
 \[user@system project] cucumber -p html_report -p bvt
 ```
 
-## Default Profile
+<b>Default Profile</b>
```
mlvandijk: Same

content/cucumber/configuration.md
```diff
@@ -80,7 +140,7 @@ With this setup, Cucumber will now use both the `bvt` profile and `html_report`
 profile, testing all Features and Scenarios tagged with `@bvt`, along with the
 progress output and HTML output.
 
-## Preprocessing with ERB
+<b>Preprocessing with ERB</b>
```
mlvandijk: Same

content/cucumber/configuration.md
```diff
@@ -95,8 +155,8 @@ So, if you have several profiles with similar values, you might do this:
    default: <%= common %> features
    html_report: <%= common %> --format html --out=features_report.html features
    ```
-
-## Environment Variables
+      
+<b>Environment Variables</b>
```
mlvandijk: For the Ruby page, this should definitely be a header

content/tools/ruby.md
```diff
@@ -66,7 +66,7 @@ while running `rake cucumber:ok` will run the Steps within all Scenarios other t
 
 ### Using Profiles in Rake Tasks
 
-For complex Feature runs that are tested often, it is nice to save the command line arguments as [Cucumber profiles](/cucumber/configuration/#defining-profiles).
+For complex Feature runs that are tested often, it is nice to save the command line arguments as [Cucumber profiles](/cucumber/configuration#profiles).
```
mlvandijk, line 69: Sidenote: afaik the links were fine, but they still seem to work like this, so ok.

mlvandijk (commented):
Thanks for taking the time to add this information! Really appreciate it.
I do have some comments, mainly regarding keeping the documentation maintainable and consistent. Let me know if you have questions / want my help.
Also, could you please fix the merge conflict? (Merged your other PR yesterday ;) )

content/cucumber/configuration.md
```diff
@@ -1,13 +1,75 @@
...
+{{% block "java" %}}
+  Profiles are not available in Java
+{{% /block %}}
+
+{{% block "javascript" %}}
+ TODO:
+{{% /block %}}
+
+{{% block "ruby" %}}
 You can specify configuration options for Cucumber in a `cucumber.yml` or `cucumber.yaml` file.
```
luke-hill: Can we remove this, as it technically isn't true. Many people have their `cucumber.yml` in their TLD. Often because `Rakefile` is also usually located in their TLD

content/cucumber/configuration.md
```diff
@@ -22,17 +84,15 @@ want to execute with this profile.
 The example above generates two profiles:
 
 1. `html_report`, with a list of command-line options that specify new output formats, and
```
luke-hill, line 126: Qu) Do we prefer to link people to what individual args do, or specify them inline? As below we stipulate what it does, but here we don't?

luke-hill (commented):
Couple of thoughts. Happy to discuss if needs be.

content/cucumber/api.md
mpkorstanje: You'll have to do this. I'm running into heaps of semantical problems here. The commandline runner is like the JUnit runner a way to run cuucumber. But only the commandline is part of all API's. 

So I don't know where this should go.

content/cucumber/configuration.md
mpkorstanje: That is the problem. This stuff needs a header. But it is very specific to ruby and has no equivalent elsewhere.

content/cucumber/api.md
mlvandijk: Np. Will do at a later time. It's outside the scope of this PR.

content/cucumber/configuration.md
mpkorstanje: The environment variables are only used in context of the configuration file though. So it's not a general section about environment variables. Just another way to configure the configuration?

content/tools/ruby.md
mpkorstanje: It complained.

content/cucumber/configuration.md
mlvandijk: Which is why I propose to have language specific configuration pages, like here: https://docs.cucumber.io/tools/general/ (one general page, pointing to language specific pages; each language specific page can have it's own headers)

content/tools/ruby.md
mlvandijk: Hmm, must be new. Glad it works :)

mpkorstanje:
@mlvandijk I've rebased against master. Conflicts are gone.

I'd rather see that some one else sorts out the organizational problems though. 

I've made a few attempts at splitting the configuration section into 3 parts but that causes even more problems. There are quite a few incoming pages.

mlvandijk:
Had another look. Tools is a separate menu, so Config would not quite be the same. Will think of an alternative.

content/cucumber/configuration.md
mpkorstanje: I think it's better to make readers aware that the documentation isn't here yet. I'll leave the resolution up to someone else though.

content/cucumber/configuration.md
mpkorstanje: Eh. I guess 'parameter- and data table types'? Not sure how that construct works in English actually.

mlvandijk:
Some links:
https://cucumber.io/blog/2017/09/21/upgrading-to-cucumber-3 (Ruby - Parameter Type)
https://cucumber.io/blog/2017/07/26/announcing-cucumber-expressions (Ruby - Cucumber expressions)
https://github.com/cucumber/cucumber-js/blob/master/features/profiles.feature (JavaScript - Profiles)

content/cucumber/configuration.md
mlvandijk: @luke-hill - If you want to make changes to the Ruby parts, please do. @mpkorstanje and I have no opinion on this ;)

PR #175, opened 2018-06-05
Type registry configuration's api changed

kristofpeto:
Instead of cucumber.api.Configuration class, there is an interface now, with these two methods.


mpkorstanje:
Already working on it:. The problem is a bit more comphehensive:  https://github.com/cucumber/docs.cucumber.io/pull/172

mlvandijk:
Can/should this be merged into #172?

mpkorstanje:
No this can be closed. I've got everything covered I believe

PR #176, opened 2018-06-06
Remove `@Delimiter` from the documentation

mpkorstanje:
Remove `@Delimiter` from the documentation. Support was dropped in `cucumber-jvm:3.0.0`

### DIFF
--- a/content/cucumber/api.md
+++ b/content/cucumber/api.md
@@ -17,23 +17,7 @@ The number of parameters in the {{% stepdef-body %}} has to match the number of 
 ## Data Tables
 
 {{% text "java" %}}
-The simplest way to pass a `List<String>` to a step definition is to use commas:
-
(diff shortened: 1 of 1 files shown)


mlvandijk (commented):
Since there are lots of people on different versions, let's distinguish between them? I.e. up until this version you could use this; as of this version do that.

mpkorstanje:
We're not doing this for anything else though? Type registry was introduced in 3.0.0.

mlvandijk:
Imho we should, if functionality was specifically introduced in a particular version. This was discussed but not resolved afaik.

mlvandijk:
Related to #175 / #172

mlvandijk:
Also @mpkorstanje - is it no longer possible to pass a list using comma's/delimiters as of v3?

PR #207, opened 2018-07-28
Ocaml Install Instructions

cyocum:
This adds more detail to the OCaml implementation landing page specifically for installation.

### DIFF
--- a/content/installation/ocaml.md
+++ b/content/installation/ocaml.md
@@ -6,11 +6,16 @@ implementation: official
 weight: 1125
 ---
 
+[Cucumber.ml](https://github.com/cucumber/cucumber.ml) is an
+implementation written in and useful for
+[OCaml](http://www.ocaml.org/) projects.
+
 # Prerequisites
 
-There are a few complications when working with Cucumber.ml. The first
-and the most important is that Cucumber.ml delegates its Gherkin
-parsing to the
+There are a few complications when working with
+[Cucumber.ml](https://github.com/cucumber/cucumber.ml). The first and
+the most important is that Cucumber.ml delegates its Gherkin parsing
+to the
 [gherkin-c](https://github.com/cucumber/cucumber/tree/master/gherkin/c)
 implementation using OCaml's foreign function interface. This means
 that, before you can use Cucumber.ml, you will need to compile and
--- a/content/installation/ocaml.md
+++ b/content/installation/ocaml.md
@@ -35,8 +35,9 @@ Cucumber.ml code, all you need to do is:
 dune build && dune install
 ```
 
-This will install the `cucumber` package into your local Opam
-repository. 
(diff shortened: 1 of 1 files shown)


content/installation/ocaml.md
```diff
@@ -6,4 +6,49 @@ implementation: official
...
 ---
 
-Please see the [Cucumber OCaml website](https://github.com/cucumber/cucumber.ml).
+[Cucumber.ml](https://github.com/cucumber/cucumber.ml) is an
+implementation written in and useful for
+[OCaml](http://www.ocaml.org/) projects.
+
+# Prerequisites
+
+There are a few complications when working with
```
mlvandijk: * Consider switching the order here: Mention the prerequisite first, and then clarify below as to why. 
* If possible/needed, provide short steps on how to meet these prerequisites and/or check that they have been met (to prevent trouble shooting later on). In general, prevent lengthy text. (For an example of using bullit point, have a look at the 10 minute tutorial also).
* If possible, show the terminal commands inside code blocks marked as shell
* You mention Linux explicitly. Are instruction the same / different for Windows / MacOS? If relevant, could you add those?

content/installation/ocaml.md
```diff
@@ -6,4 +6,49 @@ implementation: official
...
+these installed before you can build Cucumber.ml.
+
+Once you have checked out the code and have the above prerequisites
+installed, all you need to do is:
+
+```
+dune build && dune install
+```
+
+This will install the cucumber package into your local Opam
```
mlvandijk: `Cucumber` instead of `cucumber` for consistency (please check rest of text also?)

content/installation/ocaml.md
```diff
@@ -6,4 +6,49 @@ implementation: official
...
+repository. From there you can write your step definitions in OCaml
+and run your feature files against them. For more information on how
+to use Cucumber.ml, please see the Cucumber.ml README file on the
```
mlvandijk: Please add a link to the readme here.

mlvandijk (changes requested):
Thanks for this addition! I'd really like to merge it. :)
However, I do have some comments / requests; mainly for clarity and consistency. Let me know if you have any questions and/or would like to collaborate on those!

content/installation/ocaml.md
cyocum: Thanks! This is exactly what I was looking for.

content/installation/ocaml.md
cyocum: I updated this with verbatim text as I was referring to the Opam package name.  By convention, all package names in Opam are lower case.

content/installation/ocaml.md
cyocum: I removed this text as the information is in the README on the repo so it is probably better to direct the reader there.

mlvandijk (approved):

Issue #197, opened 2018-06-15
Debugging documentation out-of-date


aslakhellesoy:
_From @cameroncooke on April 27, 2018 16:28_

## Summary

The debugging documentation on the Wiki is out of date. The `debugging.rb` code does not work with `2.4.0`.

https://github.com/cucumber/cucumber/wiki/Debugging

On running the tests one or more exceptions are thrown from the `debugging.rb` code.

## Expected Behavior

I would expect that the debugging logic not to throw it's own exceptions and perform it's intended function.

## Current Behavior

The following exception is observed in the terminal when running tests.

```bash
 uninitialized constant Rails (NameError)
 ./features/support/debugging.rb:47:in `rescue in failed'
 ./features/support/debugging.rb:41:in `failed'
```

## Steps to Reproduce (for bugs)

1. Add the `debugging.rb` file to your project following the instructions found here: https://github.com/cucumber/cucumber/wiki/Debugging
2. Run the tests using `DEBUG=1` environmental variable

## Context & Motivation

I would like to be able to debug my test code. My main motivation for debugging was to be able to get a breakpoint using `pry` when an exception is thrown in one of my tests.

## Your Environment

* Version used: 2.4.0
* Operating System and version: macOS 10.13.4

_Copied from original issue: cucumber/cucumber#376_

aslakhellesoy:
_From @mlvandijk on June 14, 2018 11:36_

Note that we have moved the documentation here: https://docs.cucumber.io/
You can find the source [here](https://github.com/cucumber/docs.cucumber.io), if anyone would like to add debugging information.

olleolleolle:
I amended the wiki page (https://github.com/cucumber/cucumber/wiki/Debugging), so that there's something until a full documentation page has been added in docs.cucumber.io.

mlvandijk:
Thanks @olleolleolle! Would you like to move the Debugging info to the docs?

aslakhellesoy:
Yes - please don't add anything to the wiki anymore, it will soon be deleted.

olleolleolle:
I'll be adding a Debugging square to the other Cucumber squares in this repo's documentation. It's my first `polyglot` article here.

olleolleolle:
Alright, now I brought in that content into a proto-article.

mlvandijk:
Thank you! We can collaborate on that and merge when sufficient. I think (hope!) Andrew might like to collaborate as well.

olleolleolle:
[...]

PR #210, opened 2018-07-31
Add Debugging article from wiki content

olleolleolle:
This PR adds a polyglot article from wiki content found in the wiki article mentioned in #197 to the site.

### DIFF
--- a/content/cucumber/debugging.md
+++ b/content/cucumber/debugging.md
@@ -1,0 +1,155 @@
+---
+title: Debugging
+subtitle: How to debug failing Cucumber steps
+polyglot:
+ - java
+ - ruby
+
+---
+
+
+# Debugging failing Cucumber steps
+
+{{% block "ruby" %}}
+(The code given below calls on [the logging facilities of Ruby on Rails](https://guides.rubyonrails.org/debugging_rails_applications.html#the-logger). If you're not using Rails, replace those calls with `puts` or `warn`.)
+
+Adding the following as the contents of `features/support/debugging.rb` can be helpful in debugging failing steps:
+
+```ruby
+# rubocop:disable Lint/Debugger
+class CucumberCounters
+  @error_counter = 0
+  @step_counter = 0
+  @screenshot_counter = 0
+  class << self
+    attr_accessor :error_counter, :step_counter, :screenshot_counter
+  end
+end
+
+# `LAUNCHY=1 cucumber` to open save screenshot after every step
+After do |scenario|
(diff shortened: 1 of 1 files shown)


content/cucumber/debugging.md
```diff
@@ -0,0 +1,155 @@
...
+end
+# rubocop:enable Lint/HandleExceptions
+
+def dismiss_nav_warning
+  execute_script "$(window).unbind('beforeunload')"
+  loop_until_jquery_inactive
+end
+
+def loop_until_jquery_inactive
+  Timeout.timeout(Capybara.default_max_wait_time) do
```
luke-hill, line 125: I'm aware this has just been copied and pasted. But this method needs replacing.

Thomas (From capybara), did a lot of work to help me remove it from site_prism.

He put an article / reference here: https://github.com/natritmeyer/site_prism/issues/257

If you want to wait an implicit wait, call something like.
```
Capybara.using_wait_time(Capybara.default_max_wait_time) do
  # method call
end
```

So I think here you'd just call `page.evaluate_script('jQuery.active').zero?` - This will re-evaluate the block until it's true.

content/cucumber/debugging.md
```diff
@@ -0,0 +1,155 @@
...
+  end
+  Rails.logger.info("[Cucumber] starting the #{@scenario_name}")
+end
+
+AfterStep do |scenario|
+  CucumberCounters.step_counter += 1
+  step = CucumberCounters.step_counter
+  file_name = format('tmp/capybara/step_%03d.png', step)
+  Rails.logger.info("[Cucumber] after step: #{@scenario_name}, step: #{step}")
+  next unless scenario.source_tag_names.include? '@intermittent'
```
luke-hill, line 101: Unsure if we're pushing this, but should we wrap method args? `include?('@intermittent')`

content/cucumber/debugging.md
```diff
@@ -0,0 +1,155 @@
...
+    Rails.logger.info("[Cucumber] Can not make screenshot of #{step}")
+  end
+end
+
+# rubocop:disable Lint/HandleExceptions
+AfterStep do
+  begin
+    execute_script "$(window).unbind('beforeunload')"
+  rescue
+  end
```
luke-hill, line 115: How about adding in `Rails.logger.error("An error was encountered and rescued")`

Maybe even with `e.stacktrace` or something?

luke-hill (changes requested):
Couple of things worth changing IMO

PR #213, opened 2018-08-04
Improved Gherkin dialects

### LINKED ISSUE #212: Can't scroll to show the full content
qq 20180804195703

Just look at the screenshot above, I'm using Chrome, but the right part didn't show and I can't scroll to them.

xtrasimplicity:
# Summary
I've restructured the Gherkin Dialects table so that it renders numerous little tables, one for each language/dialect, so as to ensure that the whole table is visible regardless of a user's Browser's zoom settings.

This resolves #212.


xtrasimplicity:
I've come up with a few different options for this, what do you think? @mlvandijk @aslakhellesoy 
[Option 1: Separate by language](https://5b65b98d792f8926e39ddc7d--cucumber.netlify.com/gherkin/reference/#spoken-languages)
[Option 2: Separate by language & vertical table](https://deploy-preview-213--cucumber.netlify.com/gherkin/reference/#spoken-languages)

Alternatively, we can add an `overflow-x: scroll` attribute via CSS to the original table, but I suspect the original table is a little too busy?

mlvandijk:
@xtrasimplicity Of these 2 options, the second seems more readable to me (less horizontal scrolling). It does make the page very long though; I'm still not sure we should have it in the docs (or maybe on a separate page?)
Tiny details: There's a comma behind the last language where there shouldn't be. Also, I'd prefer the list to be in alphabetical order of languages, making them easier to find.

xtrasimplicity:
Thanks for the feedback, @mlvandijk. I'll move it to a separate page, fix the sorting and remove the comma. :smile: 

Edit: Actually, what do you think about using [collapsible, 'accordion-styled'](https://getbootstrap.com/docs/4.0/components/collapse/#accordion-example) panels, rather than having a separate page?

mlvandijk:
Love it! (Also as having them on a separate page would probably add them to the menu, and I don't think we'd want that). Does it also have an "expand all" option? (if someone wanted to) Although that might be a nice to have.  
My only concern would be whether it would make the docs harder to maintain / more prone to bugs.

xtrasimplicity:
[All done!](https://5b66b60273f2cf62a74f76ca--cucumber.netlify.com/gherkin/reference/#spoken-languages) I haven't added an 'expand all' button, but it could be very easily done with Javascript. :+1:

mlvandijk:
Love it, thanks!

PR #217, opened 2018-08-09
Use 'end' instead of })

yrral86:
Incorrect ending of block in ruby example


olleolleolle (approved):
Thanks, this looks great!

PR #228, opened 2018-09-20
Added code block to ignore subset of scenarios

Shobika:


content/cucumber/api.md
```diff
@@ -531,9 +531,19 @@ Or changing your JUnit runner class:
...
 public class RunCucumberTest {}
 ```
-
 {{% /block %}}
 
+## Ignoring a subset of scenarios
+
+You can tell Cucumber to ignore scenarios with a particular tags:
+
+Using JUnit runner class:
```
mlvandijk: 2) Please mark this line as "java" text, so it only shows up in the Java tab: 
"Using JUnit runner class:" ->
"{{% block "javascript" %}}Using JUnit runner class:{{% /block %}}"

Fyi: You can either build the project yourself locally (check the README for how to do so) and check that any text you add doesn't show up in the wrong place for other language tabs. Alternatively, you can also find a preview of your PR under "Show all checks" -> "Details" to see how it looks.

content/cucumber/api.md
```diff
@@ -531,9 +531,19 @@ Or changing your JUnit runner class:
 @CucumberOptions(tags = "@smoke and @fast")
 public class RunCucumberTest {}
 ```
-
 {{% /block %}}
 
+## Ignoring a subset of scenarios
```
mlvandijk, line 550: Nice addition. Two things (you might not be aware of):
1) This text block should be below the Javascript and Ruby sections for "Running a subset of scenarios", otherwise it will show the Javascript/Ruby code for that section under this heading.

content/cucumber/api.md
```diff
@@ -772,6 +782,53 @@ import org.junit.runner.RunWith;
 public class RunCukesTest {
 }
 ```
+For example if you want to check whether all feature file steps has corresponding step definitions, you can specify it like this:
```
mlvandijk: "has" -> "have"

mlvandijk (changes requested):
Nice addition, thank you! Some comments (mainly related to formatting, so it will continue to look nice in all language tabs). Please let me know if you want to do those yourself, or if I should pick that up?

Shobika:
@mlvandijk updated based on the review. Please let me know if I missed something. Thank you!

aslakhellesoy (commented):
Why are you adding a paragraph that is almost identical to the one above - *Running a subset of scenarios*?

The existing paragraph uses [tag expressions](https://docs.cucumber.io/cucumber/tag-expressions/) while your new paragraph uses the legacy format.

aslakhellesoy:
I might be missing the intention of this pull request. Please add a detailed description to pull requests explaining *why* you are proposing a change and what’s currently missing.

aslakhellesoy:
@Shobika could you please elaborate on the motivation for this pull request?

Shobika:
@aslakhellesoy Actually while I was reading the documentation as an end-user, I found a section to run a subset of scenarios, which had the example of using only "AND" operator. So, thought it would be better to have a section that emphasize on only ignoring the scenarios using "NOT" operator. 

There was no links to tag expressions page from this section so as an end-user, I might not have an idea of different formats of expressions. I came to know about that, only after you sent the link. Maybe we can include the tag expressions page link in running subset scenarios section so that user might have access to different expressions examples and also we can add an example of only "NOT" expression in tag expressions page so that user knows they can use only NOT operator in tags expression. 

Please let me know your inputs

mlvandijk:
@Shobika Yes, actually, let's add a link to tag expressions from the "Running a subset of scenarios" section. Are you ok to make that change?

Shobika:
Yeah i will add that change..

mlvandijk:
Thnx @Shobika

mlvandijk (approved):

aslakhellesoy (commented):
This was using a deprecated tag expression syntax. `~@smoke` should be `not @smoke`.

Could you please correct that @mlvandijk?

mlvandijk:
Yes, will do. Was thinking of merging tag expressions in here; didnt get around to it today.

Issue #144, opened 2018-04-20
Close down / archive / redirect old doc sources


mlvandijk:
Any relevant information from wiki-pages should be added to the docs. See below for a list of wiki pages. Note that a lot of content is marked "removed" because it was either too little information and/or very outdated. If the topics are considered relevant, please provide a PR with sufficient and up-to-date information.

mattwynne:
I've set up redirects from the old sites (via Cloudflare page rules):

- `docs.cucumber.io/docs/*` (the old gitbook site's URLs) now redirects to the `docs.cucumber.io` homepage
- `cucumber.io/docs/*` redirects to the `docs.cucumber.io` homepage

I didn't add any redirects for specific URLs because pragmatism.  I don't like breaking links and at the same time I don't know how important it is to keep them. If there are any specific URLs people want redirected please let me know.

The wiki has been closed for edit, and @aslakhellesoy is going through the pages to add redirects (where applicable) to the new pages.

mlvandijk:
See also #42 (possible duplicate)

mlvandijk:
TO BE REMOVED:
- [ ] https://github.com/cucumber/cucumber/wiki/Autotest-Integration (removed)
- [ ] https://github.com/cucumber/cucumber/wiki/Console-Colours (removed)
- [ ] https://github.com/cucumber/cucumber/wiki/Cucumber-Backgrounder  (content was moved to relevant sections and a lot of it was deleted)
- [ ] https://github.com/cucumber/cucumber/wiki/Cucumber-Feature (removed)
- [ ] https://github.com/cucumber/cucumber/wiki/Fixtures (removed)
- [ ] https://github.com/cucumber/cucumber/wiki/FunFX-and-Flex (removed)
- [ ] https://github.com/cucumber/cucumber/wiki/IronRuby-and-.NET (removed)
- [ ] https://github.com/cucumber/cucumber/wiki/IronRuby-and-Mono (removed)
- [ ] https://github.com/cucumber/cucumber/wiki/Patterns (removed)
- [ ] https://github.com/cucumber/cucumber/wiki/Merb (removed)
- [ ] https://github.com/cucumber/cucumber/wiki/RDoc (removed)
- [ ] https://github.com/cucumber/cucumber/wiki/Setting-up-Selenium (removed)
- [ ] https://github.com/cucumber/cucumber/wiki/Setting-up-Steam (removed)
- [ ] https://github.com/cucumber/cucumber/wiki/Sinatra (removed)
[...]

PR #242, opened 2018-10-11
fixed default link & added appropriate links to resources #144

sridhareaswaran:
**Worked on:**
Issue #144 - Close down / archive / redirect old doc

**Changes:**
1. Removed default URL in `card.html`
2. Added appropriate URLs to resources listed in `books.md`

### DIFF
--- a/content/professional/books.md
+++ b/content/professional/books.md
@@ -11,7 +11,8 @@ We like these Books
   <div class="column is-one-third">
      {{% card 
          title="Discovery"
-         image="/img/book-discovery.png"
+         image="/img/book-discovery.png" 
+         link="https://www.goodreads.com/book/show/36126762-the-bdd-books"
      %}}
 This practical guide by Gáspár Nagy and Seb Rose explains good collaboration techniques, illustrated by concrete examples.
      {{% /card %}}
@@ -19,7 +20,8 @@ This practical guide by Gáspár Nagy and Seb Rose explains good collaboration t
   <div class="column is-one-third">
     {{% card 
         title="The Cucumber Book"
-        image="/img/book-cucumber-book.jpg"
+        image="/img/book-cucumber-book.jpg" 
+        link="https://www.goodreads.com/book/show/12409185-the-cucumber-book"
     %}}
 Matt Wynne and Aslak Hellesøy show you how to express your customers’ wild ideas as a set of clear, executable specifications that everyone on the team can read.
     {{% /card %}}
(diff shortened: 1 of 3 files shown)


content/tools/general.md
```diff
@@ -25,6 +25,10 @@ It offers several [packages](https://atom.io/packages/search?q=cucumber) you can
 [TextMate](https://macromates.com/) is a text editor for macOS.
 See the [`Cucumber.tmbundle`](https://github.com/cucumber/cucumber-tmbundle) documentation.
 
+## Tidy Gherkin
```
mlvandijk: This tool should be on the [Related Tools](http://docs.cucumber.io/tools/related-tools/) page instead of the general tools page. Could you move it there please?

themes/cucumber-hugo/layouts/shortcodes/card.html
```diff
@@ -1,4 +1,4 @@
-<a href="http://google.com">
+<a href="{{ .Get "link" }}">
   <div class="card" style="height: 100%">
```
mlvandijk, line 2: Going to a specific link per book instead of Google home page is a great idea. Just not sure about what the particular links should be

content/professional/books.md
```diff
@@ -11,23 +11,26 @@ We like these Books
   <div class="column is-one-third">
      {{% card 
          title="Discovery"
-         image="/img/book-discovery.png"
+         image="/img/book-discovery.png" 
+         link="https://www.goodreads.com/book/show/36126762-the-bdd-books"
```
mlvandijk, line 15: I like having individual links per book, rather than go to google home page. Not sure it should go to GoodReads though. @aslakhellesoy @mattwynne @sebrose - do you have an opinion on this?

mlvandijk (changes requested):
Thank you for offering improvements! Just a question for the Cucumber people regarding where to link from the books (see below) and a request to move the tool to a different tools-page. 

Also, I'm not sure how this is related to #144 which is about cleaning up the old links from the wiki..

content/professional/books.md
sridhareaswaran: Ya, I felt the same. But then the reason for choosing GoodReads is that it provides genuine user reviews, multiple vendors and formats (Physical, digital & audio) for the user to choose from. Let's hear from others too.

content/tools/general.md
sridhareaswaran: Sure, I will move that.

mlvandijk (approved):

content/professional/books.md
mattwynne: From an author's point of view, linking to the Pragmatic Programmers' page for the books that are published through them means we get a better royalty payment than if you link to Amazon. Otherwise, just do what seems most useful to the reader.

PR #250, opened 2018-11-01
[cucumber-expressions] Explain that double backslashes escape

devendra-aher:


aslakhellesoy:
Thnks for the improvement! However, this is not correct. Instead of suggesting double backslashes, could you add a paragraph below saying “You may have to escape the `\` character itself with another `\`, depending on your programming language.”

For example, with ruby and single-quoted strings, a single backslash is what you should use.

devendra-aher:
Thank you @aslakhellesoy. I'll make changes according to that.

aslakhellesoy:
Next time you get feedback on a pull request, just push additional commits to the same branch to address feedback rather than creating a new PR.

Issue #52, opened 2017-10-25
Add FAQ


mlvandijk:
Would it be helpful to add an FAQ page? I've noticed some questions have come up several times just on Slack alone.
Examples:
* cannot find jvm v2.0 => group id has changed
* message that steps are not defined even though they are, due to glue path not being configured correctly

aslakhellesoy:
Yes!!! I've felt guilty about bot having one for years so starting one would be great. Best way to handle recurring questions on ML.

aslakhellesoy:
* cucumber and excel (don't, explain why)

mlvandijk:
* skipping steps (don't, explain why)

mlvandijk:
* issues with runners / how to run

plaindocs:
If at all possible, I've a strong preference for first finding the right place for that info to live, and then linking to it in the form of an FAQ.

mlvandijk:
@plaindocs Yes, I agree. However, sometimes people don't find the documentation; it might be helpful to have a list of common problems...

mlvandijk:
Additional FAQ (which should -also- be clarified in the docs):
- Getting 'undefined steps' because glue is not found (where does it look for the glue by default, how to set where to find the glue) - so many StackOverflow questions...
- Running though CLI (how to) - see: 
* https://groups.google.com/forum/#!topic/cukes/bM_ZHHwks8U
* https://www.jeroenreijn.com/2017/10/running-cucumber-from-the-command-line.html
=> Note there is a separate issue on adding this information (#261). It should be added and then linked from the FAQ to that section.
* Taking a screenshot on fail (& explain taking screenshot every step is an antipattern)

mlvandijk:
Input for "Cucumber can't find my glue" (from @brasmusson on Slack): Cucumber-JVM will search for step definitions in the packages you told it to look in. Either you put your step definitions files in the default package where Cucumber-JVM searches, that is the package (or sub-packages) of the runner class - in your case the package `com.Create_Customer_API`, or you tell Cucumber-JVM explicitly which packages (which includes the sub-packages) to search with the `glue = {"<package>", "<package>", "<etc>"}` in the `@CucumberOptions` annotation - in your case `glue = {"Create_Customer_API"}`

mlvandijk:
[...]

PR #255, opened 2018-11-06
Add a FAQ page

mlvandijk:
Add a FAQ page with answers to frequently asked question.
This is a partial fix for isse #52; not all questions mentioned in the issue are answered atm.


content/community/faq.md
```diff
@@ -0,0 +1,149 @@
...
+By default Cucumber-JVM will search in the package (or sub-packages) of the runner class.
+You can also tell Cucumber-JVM explicitly which packages (and sub-packages) to search, with:
+ ```java
+ @CucumberOptions(glue = {"<package>", "<package>", "<etc>"})
+ public class RunCucumberTest{}
+```
+{{% /block %}}
+
+## How do I call other steps or scenarios?
+Each scenarios should be *independent*; you should be able to run them in any order or in parallel without one scenario interfering with another.
```
msmorawski: Typo, should be `Each scenario`

content/community/faq.md
```diff
@@ -0,0 +1,149 @@
...
+
+Each scenario should *test exactly one thing* so that when it fails, it fails for a clear reason. This means you wouldn't reuse one scenario inside another scenario.
+
+If your scenarios use the same or similar steps, or or perform similar actions on your system, you can extract [helper methods](#how-to-use-helper-methods) to do those things.
```
msmorawski: Repeated `or`

content/community/faq.md
```diff
@@ -0,0 +1,149 @@
...
+  // sendKeys to password fields
+
+  // click Login button
+
+}
+```
+
+You can call this method from your step definition. You could even call *several* helper methods from one step definition; for instance, if you want to login and navigate to a particular page.
+
+Ofcourse, which helper methods you'll need will depend on your application and your domain.
```
msmorawski: Should be `Of course`

content/community/faq.md
```diff
@@ -0,0 +1,149 @@
...
+# How do I share state between steps?
+{{% block "ruby,javascript" %}}
+In order to share state between steps, you can use a [world object](/cucumber/state/#world-object).
+{{% /block %}}
+
+{{% block "java,kotlin" %}}
+If you are using Cucumber on the JVM, you can use [dependency injection (DI)](/cucumber/state/#dependency-injection) to share state between steps.
+
+If your project already uses a dependency framework supported by Cucumber (and/or you are familiar with one of them), it's probably easiest to use that framework. 
+Otherwise, Picocontainer is the most light weight framework you can use.
```
msmorawski, line 148: Maybe should be replaced by `lightest weight`?

mlvandijk: Hmm, not sure. Using "light weight" as an expression here.

mlvandijk: Fixed the typos though, thanks for pointing them out!

PR #265, opened 2018-11-17
[Java] Use consistent package and runner names in example

mpkorstanje:
Renamed `RunCukesTest` to `RunCucumberTest`. While Cukes are cute and informal using Cucumber consistently makes it much easier to find what you're looking for.

And as pointed out by a co-worker `mypackage` has a certain sophmoric humour to it.


olleolleolle (approved):
This is easier to understand, too.

Issue #262, opened 2018-11-13
Add "Rule" keyword to Gherkin documentation


mlvandijk:
See grammar definition in https://github.com/cucumber/cucumber/blob/master/gherkin/gherkin.berp
Also: https://github.com/cucumber/cucumber/issues/453

Note: should be added when Gherkin 6 is actually used in any/all Cucumber implementations.

TomerPacific:
@mlvandijk ,
I would like to help with this issue. Is it only about adding the word "Rule" when necessary?

mlvandijk:
Thanks @TomerPacific, that would be great. 
Yes, we should add the "Rule" keyword and relevant information about this keyword (see links).

TomerPacific:
Opened a [PR](https://github.com/cucumber/docs.cucumber.io/pull/270)

mlvandijk:
Looks like Gherkin 6 is not implemented yet in Cucumber-jvm:
the [pom.xml](https://github.com/cucumber/cucumber-jvm/blob/master/pom.xml) mentions `<gherkin.version>5.1.0</gherkin.version>` and there's an open [PR](https://github.com/cucumber/cucumber-jvm/pull/1402) to use Gherkin 6.

Cucumber-js [package.json](https://github.com/cucumber/cucumber-js/blob/master/package.json) mentions `"gherkin": "^5.0.0",`

Cucumber-ruby [cucumber.gemspec](https://github.com/cucumber/cucumber-ruby/blob/master/cucumber.gemspec) (if that's the Ruby equivalent of a pom file) does mention `  s.add_dependency 'gherkin', '~> 6.0'`.

So it looks like it's implemented for Ruby, so we should add it to the documentation. However, we should make clear it is not yet implemented in jvm/js, for which there are two options:
1. Put all text regarding `Rule` keyword inside "ruby" text blocks
2. Add a *Note* that Gherkin 6 / `Rule` keyword has not yet been implemented for jvm/js.

Personally I think option 1 would be a safer option (less confusion), but might be a bit more work. What do you think @TomerPacific , @jenisys ?

jenisys:
You can always wait until the last one us on-board. But that is composition problem that will not scale the more implementations you get. 

IMHO a better approach is to document per implementation which Gherkin version it supports. But clarify that with Aslak and/or Matt first. 

Ciao,
Jens

mlvandijk:
Closed by cucumber/common#271

PR #271, opened 2018-12-16
#262: Provide description of Rule keyword/idiom

jenisys:
Provides the description for the `Rule` keyword / idiom, up to now missing in the Gherkin reference documentation. The `Rule` keyword is supported since Gherkin v6.

### DIFF
--- a/content/gherkin/reference.md
+++ b/content/gherkin/reference.md
@@ -427,7 +427,6 @@ Cucumber provides a rich API for manipulating tables from within step definition
 See the [Data Table API reference](https://github.com/cucumber/cucumber/tree/master/datatable) reference for
 more details.
 
-
 # Spoken Languages
 
 The language you choose for Gherkin should be the same language your users and
--- a/content/gherkin/reference.md
+++ b/content/gherkin/reference.md
@@ -333,32 +333,12 @@ You can also use parameters in [multiline step arguments](#step-arguments).
 
 ## Rule
 
-The `Rule` entity is a rather new idiom in Gherkin (since Gherkin v6).
+The `Rule` keyword has been added in Gherkin v6. Note that Gherkin 6 has not yet been incorporated into all implementation of Cucumber.
 Each `Rule` is intended to represent one *business rule* that should be implemented.
-A `Rule` is used to group together several `Example` (aka: `Scenario`) / `Scenario Outline`
+A `Rule` is used to group together several scenarios 
 that belong to this *business rule*. In addition, a `Rule` may contain a `Background` section.
 
-Rule parent:
-
-- `Feature` (cardinality: one)
-
-Rule children:
-
-- `Description` (cardinality: optional, `0 .. 1`)
-- `Background` (cardinality: optional, `0 .. 1`)
-- `Example`/`Scenario` (cardinality: many, `0 .. N`)
-- `Scenario Outline`/`Scenario Template` (cardinality: many, `0 .. N`)
-
-Rule constraints:
-
-- A `Rule` should have at least one child (that is neither `Description` nor `Background`).
-
-Rule supports:
-
-- `Tags`: Allows to select a `Rule`(s) as part of a runnable set for a test runner.
-- Grouping: Of `Example`/`Scenario`, `Scenario Outline` (and `Background`)
-
-Example:
+For example:
 
 ```gherkin
 # -- FILE: features/gherkin.rule_example.feature
--- a/content/gherkin/reference.md
+++ b/content/gherkin/reference.md
@@ -39,12 +39,12 @@ Each line that isn't a blank line has to start with a Gherkin *keyword*, followe
 The primary keywords are:
 
 - `Feature`
+- `Rule`
 - `Example` (or `Scenario`)
 - `Given`, `When`, `Then`, `And`, `But`  (steps)
 - `Background`
 - `Scenario Outline` (or `Scenario Template`)
 - `Examples`
-- `Rule`
 
 There are a few secondary keywords as well:
 
--- a/content/gherkin/reference.md
+++ b/content/gherkin/reference.md
@@ -39,7 +39,7 @@ Each line that isn't a blank line has to start with a Gherkin *keyword*, followe
 The primary keywords are:
 
 - `Feature`
-- `Rule`
+- `Rule` (as of Gherkin 6)
 - `Example` (or `Scenario`)
 - `Given`, `When`, `Then`, `And`, `But`  (steps)
 - `Background`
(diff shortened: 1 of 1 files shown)


content/gherkin/reference.md
```diff
@@ -44,6 +44,7 @@ The primary keywords are:
 - `Background`
 - `Scenario Outline` (or `Scenario Template`)
 - `Examples`
+- `Rule`
```
mlvandijk: I'd prefer to have the `Rule` higher up in the list (see @TomerPacific 's PR)

content/gherkin/reference.md
```diff
@@ -330,6 +331,60 @@ to match the step against a step definition.
 
 You can also use parameters in [multiline step arguments](#step-arguments).
 
+## Rule
+
+The `Rule` entity is a rather new idiom in Gherkin (since Gherkin v6).
+Each `Rule` is intended to represent one *business rule* that should be implemented.
+A `Rule` is used to group together several `Example` (aka: `Scenario`) / `Scenario Outline`
```
mlvandijk: Could you replace "`Example` (aka: `Scenario`) / `Scenario Outline`" with "scenarios" - for readability.

content/gherkin/reference.md
```diff
@@ -330,6 +331,60 @@ to match the step against a step definition.
...
+that belong to this *business rule*. In addition, a `Rule` may contain a `Background` section.
+
+Rule parent:
+
+- `Feature` (cardinality: one)
```
mlvandijk: What do you mean with "(cardinality: one)"? A `Rule` can belong to only one `Feature`?

content/gherkin/reference.md
```diff
@@ -330,6 +331,60 @@ to match the step against a step definition.
...
+
+Rule children:
+
+- `Description` (cardinality: optional, `0 .. 1`)
```
mlvandijk: Do we need to mention the cardinality? Does that clarify things or would it be a possible source of confusion?

content/gherkin/reference.md
```diff
@@ -330,6 +331,60 @@ to match the step against a step definition.
...
+- `Background` (cardinality: optional, `0 .. 1`)
+- `Example`/`Scenario` (cardinality: many, `0 .. N`)
+- `Scenario Outline`/`Scenario Template` (cardinality: many, `0 .. N`)
+
+Rule constraints:
+
+- A `Rule` should have at least one child (that is neither `Description` nor `Background`).
```
mlvandijk: Is this checked by the parser? What if I have a `Rule` but no scenario/example?

content/gherkin/reference.md
```diff
@@ -330,6 +331,60 @@ to match the step against a step definition.
...
+
+Rule supports:
```
mlvandijk: I like how you add these headers (children, contraints, etc)! I'm not sure this particular one adds value though; as this information is also provided elsewhere afaik. (Although maybe check that Rule is mentioned under Tags). 
Now I'm wondering if we should have similar headers for all the other keywords...?
I also like that you put an example!

content/gherkin/reference.md
```diff
@@ -372,6 +427,7 @@ Cucumber provides a rich API for manipulating tables from within step definition
 See the [Data Table API reference](https://github.com/cucumber/cucumber/tree/master/datatable) reference for
 more details.
 
+
```
mlvandijk: Extra white line, remove?

content/gherkin/reference.md
```diff
@@ -330,6 +331,60 @@ to match the step against a step definition.
 
 You can also use parameters in [multiline step arguments](#step-arguments).
 
+## Rule
+
+The `Rule` entity is a rather new idiom in Gherkin (since Gherkin v6).
```
mlvandijk: Suggestion: "As of Gherkin 6, a new keyword `Rule` has been added." 

**Note:** I'm not sure Gherkin 6 has been implemented for all languages yet. Afaik it is still pending for Cucumber-jvm (and I don't use Js / Ruby personally).

mlvandijk (commented):
Hi @jenisys - thanks for taking the time to do this PR. Too bad we have 2 on the same topic; let's see if we can merge your efforts with @TomerPacific 's :)

I have a few comments/suggestions, and maybe @TomerPacific can also have a look.

content/gherkin/reference.md
jenisys: Yes, I could. 

But you have no heading on this page that says "Scenario", only "Example" because it is the new, preferred keyword (as I understand it from the rest of the description on this page).
Therefore, the docs would be no longer self-explaining for a beginner in my opinion.

@mlvandijk 
Just decide what you think is best (and I make the change).

content/gherkin/reference.md
jenisys: Corrected, I put it under `Feature` item.

NOTE: 
The old ordering represented the heading ordering on this page.
The reason I put the section w/ `Rule` further down is because it is a new addition and the other sections are more important (IMHO).

content/gherkin/reference.md
jenisys: Exactly, a one-to-one relationship has the cardinality "one" (in the direction towards the `Feature`). Cardinality "one" means if a `Rule` is  used, then the Rule must always be below a `Feature`.

I assume, you know what an UML class diagram is ?!?
(and the cardinality for a relation between classes or objects).

content/gherkin/reference.md
jenisys: You are basically describing a grammar in your own words.
You can leave it out. But then it is no longer a (readable) specification. 
Meaning, how do you understand how man grammar-elements may occur.

@mlvandijk 
Just decide.

content/gherkin/reference.md
jenisys: You basically have a composition problem. Composition problems often do not scale.
The more Cucumber implementations you get, the worse you are. Meaning, one implementation may block you forever.

A better approach would be to document that the Javascript implementation describes which Gherkin version it supports (or there is a table for all implementations).

MAYBE: @aslakhellesoy and @mattwynne should probably provide some feedback what is best.

Otherwise, decide what you think is best.

content/gherkin/reference.md
jenisys: Therefore, I use the word "should".  A `Rule` w/o `Example` makes no sense (IMHO) but is valid Gherkin.

Currently, I am not running the Gherkin parser. Therefore, I do not know.
A while ago, there was a discussion about a `Gherkin linter`. But I am not quite sure how it ended. The `linter` should provide such warnings (if the Gherkin parser does not).

content/gherkin/reference.md
jenisys: Fixed.

PR #290, opened 2019-01-16
More idiomatic Kotlin in examples plus fixed compilation error

spityi:
1. When a class is empty the open and close curly braces are not mandatory.
2. Compared to the java example, the properties of StepDefs should also be private.
3. I think the deleted lines were redundant.
4. Fixed compilation error with "this" qualifier.
(Test function names can be more readable, e.g. `I should be told` instead of i_should_be_told, but maybe modifying the generated code is not advised for this example)


mlvandijk (approved):
Thanks a lot for this @spityi !

PR #293, opened 2019-01-23
Fix broken ordering for the Gradle steps

streetturtle:


mpkorstanje:
Other then whitespace I can't see any changes.

streetturtle:
Ordered list in markdown brakes and starts from 1 if text in paragraph starts from the beginning of the line. Indenting the content allows numbering continues. Here is similar question from SO: https://stackoverflow.com/questions/18088955/markdown-continue-numbered-list

mlvandijk:
Thanks, that was driving me mad.

mlvandijk:
Hop into the #docs channel on Slack if you're up for helping out some more @streetturtle !

Issue #284, opened 2019-01-05
Additions to FAQ Page


mlvandijk:
This issue contains several FAQ that we would like to add to the [FAQ page](https://docs.cucumber.io/community/faq/)

There are several in this issue description, separated by lines.
There are several in the comments. 
They will need to be written in a similar format as currently on the FAQ page. Ideally they should answer the question and point to the right place in the docs for more information. 

**== Help wanted ==**
- Partial PR's are fine too! (i.e. adding only one of the questions/topics from this issue to the FAQ page)
- Feel free to add topics / info here that you think should be in the FAQ :)
- Please ask for clarification if anything is unclear. 
==========================

** Use of colons **
From gitter: Jeff Johnson @jeffcjohnson Jan 16 20:45
When writing features, if a newbie like me puts a colon after Given, When or Then, it doesn't complain but doesn't do anything either. It took a while to figure out why my test was being ignored. 
It would be helpful to clarify somewhere in the docs, possibly the Gherkin page. 
UPDATE: In progresss, see https://github.com/cucumber/docs/pull/691

mlvandijk:
~~**Arity mismatch**
`cucumber.runtime.CucumberException: Arity mismatch:` indicates that the step does not provide the arguments needed for the step definition.~~ (DONE in #295)

TomerPacific:
@mlvandijk ,
what exactly would you like added to the FAQ page?

mlvandijk:
@TomerPacific I would like to add additional FAQs to the FAQs page, in a similar style as the ones we have now. Ideally they should answer the question and point to the right place in the docs for more information. You can start with one of these questions mentioned if you are more comfortable with that.  We can collaborate on it through review process if you want (or join us on Slack in the #docs channel if you'd like to discuss more). Feel free to add additional questions + answers you think should be added either here in this issue, or through PR.

TomerPacific:
@mlvandijk , I have created a [PR](https://github.com/cucumber/docs.cucumber.io/pull/295) which is a work in progress so that we can collaborate and identify what else needs to be added to the FAQ. Currently, I have only added the statement regarding arity mismatch.

olleolleolle:
> **Arity mismatch**
> `cucumber.runtime.CucumberException: Arity mismatch:` indicates that the step does not provide the arguments needed for the step definition.

Add this FAQ explanation for Ruby, too.

Hat-tip: @luke-hill 

Update: **DONE**

mlvandijk:
[...]

PR #295, opened 2019-01-26
FAQ: Add Arity mismatch exception explanation (see #284)

TomerPacific:
Added to the bottom of the FAQ page a clause indicating what is the meaning of Arity mismatch.

Fixes part of the issues related to #284


luke-hill:
For reference this also occurs in ruby, so might want to add that in? It occurs in all languages I believe that take regex captures.

mlvandijk:
@luke-hill This PR adds it for all languages. The exact name of the exception in Java is shown only for Java/Kotlin.

PR #299, opened 2019-03-01
Installing Java without a dependency manager

intranetmouse:
Add some instructions for adding jars manually.

### DIFF
--- a/content/installation/java.md
+++ b/content/installation/java.md
@@ -70,6 +70,10 @@ repositories {
 
 You can now run Cucumber [from the command line](/cucumber/api/#from-the-command-line) to execute by [adding a cucumber task](/tools/java#gradle) to `build.gradle`.
 
+# Using Cucumber without a dependency manager
+
+If you have not migrated to using a dependency manager, you can still download the jars separately from the Maven central repository.  Simply go to [the Cucumber section of the Maven central repository](http://repo1.maven.org/maven2/info/cukes/) and download the latest version of `cucumber-core`, `cucumber-java`, `cucumber-jvm-deps`, and `gherkin`.  Then add them to your classpath.  If you use it through JUnit, you may also want to add `cucumber-junit` and `cucumber-html`.  Be careful that the newest version may not be at the bottom of the file list, as 10 sorts alphabetically before 9.  Also be sure to download the same version for each of the Cucumber jars.
+
 # JUnit-integration
 
 It is also possible to use [cucumber-junit](/cucumber/api/#junit) to run your Cucumber test suite.


olleolleolle (commented):
Thanks for adding this!

I have some small input on the phrasing:

Can we avoid saying “simply”? Dropping that word can change the tone without watering down the sentence. 

If we refer to them as “JAR files” instead of “jars”, our language includes readers not far into their Java jargon journey. 

This is a solid addition, I think.

content/installation/java.md
```diff
@@ -70,6 +70,10 @@ repositories {
 
 You can now run Cucumber [from the command line](/cucumber/api/#from-the-command-line) to execute by [adding a cucumber task](/tools/java#gradle) to `build.gradle`.
 
+# Using Cucumber without a dependency manager
+
+If you have not migrated to using a dependency manager, you can still download the jars separately from the Maven central repository.  Simply go to [the Cucumber section of the Maven central repository](http://repo1.maven.org/maven2/info/cukes/) and download the latest version of `cucumber-core`, `cucumber-java`, `cucumber-jvm-deps`, and `gherkin`.  Then add them to your classpath.  If you use it through JUnit, you may also want to add `cucumber-junit` and `cucumber-html`.  Be careful that the newest version may not be at the bottom of the file list, as 10 sorts alphabetically before 9.  Also be sure to download the same version for each of the Cucumber jars.
```
mpkorstanje, line 75: I don't think it is good to include this information here. It his highly perishable and not quite accurate. We publish using maven to avoid exactly this problem. 

Also cucumber has changed the `groupId` to [`io.cucumber`](https://search.maven.org/search?q=io.cucumber) and the dependency tree is more complicated now.

olleolleolle: Ah-ha!

intranetmouse: @mpkorstanje, While a dependency manager has definite technical advantages, there are numerous reasons one might not have changed over to using one (too difficult to migrate existing build, philosophical differences to some aspect of how they work, etc.).  I believe it is fair to say maven is required to be a developer on the Cucumber code.  However, making such requirements for end-users may alienate and possibly drive away potential users.  I think driving adoption of BDD through Cucumber is the prime directive, not the adoption of a particular build tool.

Perhaps there is a way (or many ways!) of making the text more generic and useful and less perishable, but I believe there is still a need for supporting users that don't use the same build tool.

mpkorstanje: Cucumber is an open source project. Everything it contains, the code, but also the documentation needs to be be maintained. This maintenance is currently done by volunteers.  Limiting the scope of the documentation to support only standard tooling and processes reduces that burden.

I can appreciate that not every project uses Maven. Fortunately Maven is not required. Other tools also provide transitive dependency management. The Maven Central website lists the exact syntax for these different tools in the side by bar. Even Ant is covered by Apache Ivy.
As such I am unconvinced by your argument. I feel that projects, or rather organizations should bear the burden and the cost of choosing not to use industry standard solutions.

This doesn't mean I don't appreciate your intent to contribute. It is very welcome, but I do request that you look at the project as a whole and towards the long term.

intranetmouse: Fair enough.  I will go ahead and reject/close this pull request based on your argument.  Thanks for your patience.

Perhaps if I get to that point, I may submit a short section on ant+ivy integration.

The only other thing that might help non-dependency tool users would be an old-school ZIP distribution of the project, which is apparently [doable in Maven](https://stackoverflow.com/a/33214362); based on my simplified understanding, the build probably would not need to be updated for every dependency change once something like this was implemented.

intranetmouse:
Closing based on feedback from mpkorstanje.

content/installation/java.md
mpkorstanje: That is certainly doable. Do you have a personal interest in such an artifact and would you be interested in maintaining it?

intranetmouse: I downloaded the code and looked at the build to see if I could figure out how such a task might be done.  While it would be nice to be able to help out, I simply do not have anywhere near enough understanding of Maven to take it on.

mlvandijk:
Thanks @mpkorstanje - fully agree ;)

PR #308, opened 2019-03-17
Added project kosher

cbush06:


aslakhellesoy:
I wrote this in 2010 or so: http://aslakhellesoy.com/post/11055981222/the-training-wheels-came-off

Nothing has changed - we should not promote reusable step definitions on the Cucumber website.

PR #317, opened 2019-03-28
Update java.md to reflect recommended way of adding a Gradle dependency and task

DragosDumitrache:
Existing code snippets no longer work with Gradle 5, as `testRuntime` has been deprecated. Updating the documentation to reflect the recommended way of doing things.

PR #318, opened 2019-03-28
Update installation instruction for Gradle version

DragosDumitrache:
Updating installation instruction for Gradle 4 and 5

### DIFF
--- a/content/installation/java.md
+++ b/content/installation/java.md
@@ -44,6 +44,7 @@ You can now run Cucumber [from the command line](/cucumber/api/#from-the-command
 If you are going to use the lambda expressions API (Java 8) to write the step
 definitions, add the following dependencies to `build.gradle`:
 
+## Gradle 4.10.3 or older
 ```
 dependencies {
     testCompile 'io.cucumber:cucumber-java8:{{% version "cucumberjvm" %}}'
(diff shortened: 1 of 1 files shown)


content/installation/java.md
```diff
@@ -44,6 +44,7 @@ You can now run Cucumber [from the command line](/cucumber/api/#from-the-command
 If you are going to use the lambda expressions API (Java 8) to write the step
 definitions, add the following dependencies to `build.gradle`:
 
+## Gradle 4.10.3 or older
```
mlvandijk: Maybe also don't use headers here, but versions in **bold**, consistent with your other PR (#317)

mlvandijk (commented):
Thanks for this @DragosDumitrache! Small change request to make it consistent with the other PR if you don't mind..

content/installation/java.md
DragosDumitrache: Sure thing. I only thought about using headers as it was easier to quickly navigate to the appropriate instructions based on the Gradle versions one uses from a table of contents, or similar.

I've also had a look at other places where updates might be required, but nothing immediately apparent

mlvandijk (approved):

mlvandijk:
Thanks again @DragosDumitrache!

PR #324, opened 2019-04-02
Fill Cucumber-Rust page with information from website

humb1t:


content/installation/rust.md
```diff
@@ -6,4 +6,147 @@ implementation: unofficial
 weight: 1200
 ---
 
-Please see the [Cucumber-Rust website](https://github.com/bbqsrc/cucumber-rust).
+## Cucmber-Rust
```
mlvandijk: Build fails because: "First header must be a h1, but it was <h2".
Could you please change your headers to start with H1, and so on?

content/installation/rust.md
```diff
@@ -6,4 +6,147 @@ implementation: unofficial
...
+
+The full gamut of Cucumber's Gherkin language is implemented by the 
+[gherkin-rust](https://github.com/bbqsrc/gherkin-rust) project. Most features of the Gherkin 
+language are parsed already and accessible via the relevant structs.
+
+### License
+
+This project is licensed under either of
+
+ * Apache License, Version 2.0 ([LICENSE-APACHE](LICENSE-APACHE) or http://www.apache.org/licenses/LICENSE-2.0)
```
mlvandijk: Please make sure to link to an actual page/URL here, as the build is failing on these links not working.

mlvandijk (changes requested):
Thank you for adding this to the docs. A few technical comments to help pass the build. Please let me know if you're ok with making these changes or have any questions. 
I'll also have a look at the content itself.

content/installation/rust.md
```diff
@@ -6,4 +6,147 @@ implementation: unofficial
...
-Please see the [Cucumber-Rust website](https://github.com/bbqsrc/cucumber-rust).
+## Cucmber-Rust
+
+Cucumber-Rust is an unofficial implementation of the Cucumber testing framework for Rust. Fully native, no external test runners or dependencies.
+
+### Usage
+
+Create a directory called `tests/` in your project root and create a test target of your choice. In this example we will name it `cucumber.rs`.
+
+Add this to your `Cargo.toml`:
```
luke-hill: Has this file been auto-generated somehow above that I've missed....

For example in ruby you can do `cucumber --init`

content/installation/rust.md
```diff
@@ -6,4 +6,147 @@ implementation: unofficial
...
+cucumber_rust = "^0.5.1"
+```
+
+If using stable Rust, you need to disable the `nightly` feature:
+
+```
+cucumber_rust = { version = "^0.5.1", default-features = false }
+```
+
+Create a directory called `features/` and put a feature file in it named something like `example.feature`. It might look like:
```
mlvandijk:
```suggestion
Create a directory called `features/` and place your feature file there. For example, a file named `example.feature`, which might look like this:
```

content/installation/rust.md
```diff
@@ -6,4 +6,147 @@ implementation: unofficial
...
+
+  Scenario: An example scenario
+    Given I am trying out Cucumber
+    When I consider what I am doing
+    Then I am interested in ATDD
+    And we can implement rules with regex
+
+```
+
+And here's an example of implementing those steps using our `tests/cucumber.rs` file:
```
mlvandijk:
```suggestion
Here's an example of how to implement those steps using our `tests/cucumber.rs` file:
```

content/installation/rust.md
```diff
@@ -6,4 +6,147 @@ implementation: unofficial
 weight: 1200
 ---
 
-Please see the [Cucumber-Rust website](https://github.com/bbqsrc/cucumber-rust).
+## Cucmber-Rust
+
+Cucumber-Rust is an unofficial implementation of the Cucumber testing framework for Rust. Fully native, no external test runners or dependencies.
```
mlvandijk:
```suggestion
Cucumber-Rust is an unofficial implementation of the Cucumber testing framework for Rust. The implementation is fully native. It doesn't use any external test runners or dependencies.
```

content/installation/rust.md
```diff
@@ -6,4 +6,147 @@ implementation: unofficial
...
+```toml
+[[test]]
+name = "cucumber"
+harness = false # Allows Cucumber to print output instead of libtest
+
+[dev-dependencies]
+cucumber_rust = "^0.5.1"
+```
+
+If using stable Rust, you need to disable the `nightly` feature:
```
mlvandijk:
```suggestion
If you are using stable Rust, you need to disable the nightly feature:
```

content/installation/rust.md
```diff
@@ -6,4 +6,147 @@ implementation: unofficial
...
+
+Create a directory called `features/` and put a feature file in it named something like `example.feature`. It might look like:
+
+```gherkin
+Feature: Example feature
+
+  Scenario: An example scenario
+    Given I am trying out Cucumber
+    When I consider what I am doing
+    Then I am interested in ATDD
```
mlvandijk, line 40: Is there a reason you are referring to ATDD instead of BDD?

content/installation/rust.md
mlvandijk: And tbh I don't know if licence needs to be mentioned here (assuming it's available in the project).

mlvandijk (changes requested):

content/installation/rust.md
humb1t: This is example from official repo

content/installation/rust.md
humb1t: Fixed

content/installation/rust.md
humb1t: Fixed

content/installation/rust.md
humb1t: In this case, you should do it by hands. I changed description a little to make it more clear.

humb1t:
@mlvandijk  @luke-hill - thanks for reviews, please look at the new version and give me feedback on what should be done next.

content/installation/rust.md
mlvandijk: OK, thanks :)

mlvandijk:
Content wise it looks fine to me. (Did some very small edits).

Maintenance wise: depending on how often a new version of Cucumber-Rust is released, you could add the `cucumber-rust` version number to the [versions.yaml](https://github.com/cucumber/docs.cucumber.io/blob/master/data/versions.yaml) and use that on the page (see the Cucumber-jvm page for how to do that, or let us know if you need help).

aslakhellesoy (changes requested):
For software that isn’t maintained by the Cucumber core team I think we should limit documentation to a couple of paragraphs.

We’re happy to link to external web pages.

This looks like it’s too elaborate. Nobody from the core team would be able to maintain it, and when it goes out of date we’d have to delete it.

Can you shorten this to a couple of paragraphs and remove code examples please?

aslakhellesoy (commented):
This PR seems to copy the docs from the original repo. I’m not sure what the motivation is, but I suggest not merging this and leaving the page like it is.

PR #331, opened 2019-04-08
Ignoring a subset of scenarios

narendracubicit:
to ignore a scenario we can also use ~ operator.


xtrasimplicity:
This has been deprecated in favour of `not`, if I'm not mistaken?

Edit: yep, deprecated, as shown in https://github.com/cucumber/cucumber-rails/issues/346. Thanks, though!

narendracubicit:
Yeah, Thank you for the update.

PR #335, opened 2019-04-21
Add navigation to docs (not homepage)

mxygem:
Trying to prevent Hotel California'ing folks on the docs page from the rest of our site's pages. AKA being unable to easily get to other pages from docs. While I know that there are ways to get to those other pages already in there, this helps make that easier. Mainly, this serves to continue to help hopefully drive traffic to the services pages.(training/school)

Not doing anything fancy, just getting something that's similar to the main page and blog set up. If there are objections, happy to chat about them, let me know!

If there are none in say, a week, I'd like to merge then. Obviously happy to get it in sooner, too!

Here's what it looks like not hamburgered:
header
And hamburgered:
hamburgered

### DIFF
--- a/JSNOTES.md
+++ b/JSNOTES.md
@@ -1,0 +1,4 @@
+# Notes
+
+* Should be change the link to training under professional services away from https://docs.cucumber.io/professional/training/ and point to training directly like we are with school?
+
--- a/JSNOTES.md
+++ b/JSNOTES.md
@@ -1,4 +1,0 @@
-# Notes
-
-* Should be change the link to training under professional services away from https://docs.cucumber.io/professional/training/ and point to training directly like we are with school?
-
--- a/themes/cucumber-hugo/layouts/partials/head.html
+++ b/themes/cucumber-hugo/layouts/partials/head.html
@@ -10,4 +10,5 @@
   <link rel="stylesheet" href="/css/cucumber.css" type="text/css">
   <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
   <script src="/js/site.js"></script>
+  <link href="https://fonts.googleapis.com/css?family=Roboto|Roboto+Slab" rel="stylesheet">
 </head>
--- a/themes/cucumber-hugo/layouts/partials/navbar.html
+++ b/themes/cucumber-hugo/layouts/partials/navbar.html
@@ -13,16 +13,53 @@
 
   <div class="navbar-menu" id="navbarMenu">
     <div class="navbar-end">
-      <a class="navbar-item" href="https://github.com/cucumber/cucumber">
-        <span class="icon">
-          <i class="fab fa-github"></i>
-        </span>
-      </a>
-      <a class="navbar-item" href="https://twitter.com/cucumberbdd">
-        <span class="icon">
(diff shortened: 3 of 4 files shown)


luke-hill (approved):
Nothing major of note; spacing in css files mainly.

Quick question: Should we be maintaining these files or are they auto-generated from jekyll?

mlvandijk (commented):
Hey @js - This is a great idea. Thanks for taking the time to work on it! 
The idea looks good to me, but I'd like to make it consistent with the website's navigation.
So I'd suggest using **BOLD** and **CAPS**.
Also, I was a little confused by the 2 lines of menu at first, so I had a look at the current website: The second line of the menu is actually inside a picture. Not sure how to make this consistent; suggestions welcome! (Also, not a must have imho)
Finally, on the current site, the Cucumber logo top left already takes you (back) to the cucumber.io site and not the docs landing page, meaning there is no way to get back to the docs landing page. Not sure if this is what we want.

mlvandijk:
Also, would there be any way to use the *actual* menu from the site (in case it changes, I'd hate to have to update it in 2 places). Not sure if technically possible.

mxygem:
Hey @mlvandijk thank you for taking a look! Take a look at the way the blog's navigo is set up as it's an attempt at getting things to look close enough on a site that's set up quite differently.

I'm totally happy to get the bold/caps thing setup for the docs site and to generally get it looking more similar but AFAIK there's no way to get it using the same exact setup. One of the battles I've faced through my blog transition is that the main site is using a Squarespace theme and the blog is using a different as they're both on different platforms.

Matt told me he's fine with things looking different between sites so we've got some flexibility, my work on the blog's navigation was part of my own desire to get things more consistent.

I'm not sure how to handle the logo's navigation behavior. There are two routes that pop into mind. Clicking the logo:
1. Takes you to the 'home' of wherever you happen to be.
2. Takes you to the main website

My feeling is that the second route would be more consistent with the wider web's behavior, would be more consistent across the site(1 makes sense for the docs and blogs, but not for training), and would make our lives easier as the maintainers.

I think keeping with the second option, this emphasizes needing to have consistent navigation, or at least options for users. With this change, docs with have it's own link in navigation as 'documentation' so I guess that's act as the 'home' button in a sense?

mlvandijk:
Hi @js - thanks for the additional info. 
I agree with the navigating to home, so that's fine. (Also the "docs" menu still going to the docs landing is a good point; hadn't thought about that.)
Also, I expected for it to not be possible to share the navigation for technical reason, so that's fine too.
Personally I'd prefer the menu to look as similar as possible, so if you could make it Bold & Caps, that would be my preference.

mxygem:
Deal!

mxygem:
@mlvandijk How do you feel about changing the doc's sites fonts? The "official" font of Cucumber is Roboto, at least for a sans serif font.

mlvandijk:
I'd be in favour of using the "oficial" font. (If you know how to change it; I don't front-end)

mxygem:
Yea, definitely!

mxygem:
Now uses Roboto and uses caps. I'm not working to get it exactly like the blog for instance, though @mlvandijk if you'd prefer it closer still, let me know.

navcaps

mlvandijk (approved):

mlvandijk:
Thanks @jaysonesmith

PR #336, opened 2019-04-21
Nest pages under /docs and update as needed

mxygem:
To coincide with our trying to move away from having docs as a subdomain and use the /docs path instead, I've updated things to point there. (docs.cucumber.io will still work, it'll just redirect to /docs)


mxygem:
Hooray for netlify's build checks! :tada: Will fix :) Also, should have fully read the readme and run things locally. :sweat_smile:

mxygem:
Merged on my own as to help avoid getting incorrect links cached by folks(and google)

PR #346, opened 2019-05-02
Removed 'Background' under 'Rule' in ninjas example

jfhector:
The ninjas example, which explains 'Rule', has a 'Background' under the 'Rule'.
This is a bug in the docs as a 'Rule' can't have a 'Background' underneath it.

This PR introduces two changes:

## 1. Removed 'Background' and its content under 'Rule'

I've removed the 'Background' under the 'Rule'.

## 2. Refactored the 'Given' statement that was under 'Background'

I've also refactored the 'Given' statement that was introduced by that 'Background'.

When refactoring the 'Given' statement across the two 'Example's, I found that the second example didn't need that 'Given' statement. If I had added "Given there are 3 ninjas" from 'Background' to the second example, it would have read like this:

  Rule: There can be only One

    Example: Only One -- One alive
      Given there are 3 ninjas
      Given there is only 1 ninja alive
      Then he (or she) will live forever ;-)

I believe that the first two Given statements are redundant. So I left the 1st one (which came from 'Background') out.

### DIFF
--- a/content/docs/gherkin/reference.md
+++ b/content/docs/gherkin/reference.md
@@ -113,7 +113,7 @@ Feature: Highlander
 
     Example: Only One -- More than one alive
       Given there are 3 ninjas
-      And there are more than one ninjas alive
+      And there are more than one ninja alive
       When 2 ninjas meet, they will fight
       Then one ninja dies (but not me)
       And there is one ninja less alive
--- a/content/docs/gherkin/reference.md
+++ b/content/docs/gherkin/reference.md
@@ -111,11 +111,9 @@ Feature: Highlander
 
   Rule: There can be only One
 
-    Background:
-      Given there are 3 ninjas
-
     Example: Only One -- More than one alive
-      Given there are more than one ninjas alive
+      Given there are 3 ninjas
+      And there are more than one ninjas alive
       When 2 ninjas meet, they will fight
       Then one ninja dies (but not me)
       And there is one ninja less alive


jfhector:
Here's why I didn't add the 'Given' statement from the 'Background' to the second example. If I had, the second example would have read like this:

  Rule: There can be only One

    Example: Only One -- One alive
      Given there are 3 ninjas
      Given there is only 1 ninja alive
      Then he (or she) will live forever ;-)

I believe that the first two Given statements are redundant. So I left the 1st one (which comes from 'Background') out.

mlvandijk:
Looks good. Thanks for your help!

aslakhellesoy:
Hi @jfhector,

Thanks for your making your first contribution to Cucumber, and welcome to the Cucumber committers team! You can now push directly to this repo and all other repos under the cucumber organization! 🍾

In return for this generous offer we hope you will:

* ✅ Continue to use branches and pull requests. When someone on the core team approves a pull request (yours or someone else's), you're welcome to merge it yourself.
* 💚 Commit to setting a good example by following and upholding our [code of conduct](https://cucumber.io/conduct) in your interactions with other collaborators and users.
* 💬 Join the [community Slack channel](http://cucumber.io/support#slack) to meet the rest of the team and make yourself at home.
* ℹ️ Don't feel obliged to help, just do what you can if you have the time and the energy.
* 🙋 Ask if you need anything. We're looking for feedback about how to make the project more welcoming, so please tell us!

On behalf of the Cucumber core team,
Aslak Hellesøy
Creator of Cucumber

Issue #298, opened 2019-02-27
Document how to run tests in parallel


mlvandijk:
We are getting some questions in Slack on how to run Cucumber-jvm in parallel. This is possible as of v4(?). We should document how to do this correctly. 

Some info: https://cucumber.io/blog/2018/09/24/announcing-cucumber-jvm-4-0-0

http://grasshopper.tech/283/

Note: As I have not had use for running in parallel, I am not familiar with this. Help very much wanted!

mlvandijk:
Denis Sln   [Feb 19th at 11:55]
Hi people!
I'm trying to find best parallel execution solution using cucumber4-java-testng. Currently, I was able to execute scenarios in parallel threads, but I want to try running each scenario in separate JVM to compare performance, however with out-of-box solution I am unable to fork new JVM instances. Tried both surefire and failsafe, no luck? Is it possible to execute scenarios in parallel JVM-s using just a single Cucumber runner and no other 3rd party plug-ins (edited)

16 replies

gabel   [8 days ago]
When you tried surefire and failsafe, did you use `threadCount` or `forkCount`? `forkCount` is the parameter that actually runs a test in a separate JVM. https://maven.apache.org/surefire/maven-failsafe-plugin/examples/fork-options-and-parallel-execution.html

Denis Sln   [8 days ago]
@gabel I tried both, also used combinations. Went through that maven references several times this weekend)

gabel   [8 days ago]
That's odd. I'm in the cucumber 2 world still, so I use an extra maven plugin which creates the runner files properly and reassembles all the reports. https://github.com/temyers/cucumber-jvm-parallel-plugin
temyers/cucumber-jvm-parallel-plugin
Maven plugin to help running Cucumber features in parallel
Stars
122
Language
Java
temyers/cucumber-jvm-parallel-pluginJan 30th, 2015Added by GitHub

gabel   [8 days ago]
But I think that same support is now built into cucumber 4, so I'm not really sure. what the issue is then.

Denis Sln   [8 days ago]
That's what I heard as well and trying to figure out how it works. I switched from Cucumber 1.2.x version. Used it together with Cucables plug-in seems to be doing similar job to the one you are referencing. Generating separate runners

Denis Sln   [8 days ago]
yeah me neither and no warnings or errors

Denis Sln   [8 days ago]
it's just when I'm adding
<systemPropertyVariables>            <databaseSchema>MY_TEST_SCHEMA_${surefire.forkNumber}</databaseSchema>
[...]

PR #347, opened 2019-05-02
Parallel execution docs

grasshopper7:
Added for [Document how to run tests in parallel](https://github.com/cucumber/docs.cucumber.io/issues/298) issue. Added two new files - parallel-execution.md and parallel-timeline-report.png.

For java only but I think with some changes could be adapted for kotlin. Have used Maven only for dependencies and execution.  For Grade need to figure it out.

Maybe need a repository to store the source code files. Though it is pretty bare bones basic stuff.


mlvandijk:
Thanks @grasshopper7 - I'll have to have a look when I have a little more time, so I can try this out.

mpkorstanje:
Quick remarks, sorry for being terse, on mobile:

1. Don't use `M` versions. They're release mile stones and still unstable.
2. Listing all versions of all jars is unmaintainable and prone to change.
3. Don't use a `feature` folder. Use java conventions and put the features in the same package as the runner. E.g. `com.example.app.RunCucumberTest` will automatically find all features in `src/test/resources/com/example/app`. This also removes some config.

mpkorstanje:
4. Same for the glue. If its in a descendant package from the runner glue need not be configured. Using these conventions helps keep the docs simple.

mpkorstanje:
5. Use Maven conventions and make sure the runner class ends with `Test.java`. E.g. `RunCucumberTest.java`. Maven will pick this up automatically. So then the config that references `Runner.java` can also be removed.

mpkorstanje:
6. I would leave out the failsafe config all together. Just link the relevant maven documentation. This is Cucumber docs, not maven docs.

mpkorstanje:
7. Looks pretty good!

aslakhellesoy:
Can you use hugo tags for versions? See the general cucumber-jvm install doc for examples

grasshopper7:
@mpkorstanje  Have updated the changes except for point 2.

Point 2 - _Listing all versions of all jars is unmaintainable and prone to change._ - Yep knew this one is gonna be a mess. Thought about mentioning latest version but the issue is with 4.3.0, it gets gherkin 5.1 and cucumber-expression 6.6.2, which are are not latest versions in repo.  Maybe mention something like - "Use the IDE to figure out transitive dependency versions of cucumber-java8 to add in classpath". Any thoughts?

Point 6 - _I would leave out the failsafe config all together._ - Think it makes things clearer regarding failsafe configurations. Have kept the junit config and referred it from the testng part. Added links to community page of tools containing maven execution details inside the docs.

Point 3,4,5 - Done simplifies things a lot.

Point 1 - _Don't use M versions. They're release mile stones and still unstable_. - Updated.

grasshopper7:
@aslakhellesoy 
> Can you use hugo tags for versions? See the general cucumber-jvm install doc for examples

I added the version tags but it is showing **4.2.6** instead of latest **4.3.0**. Any pointers how to fix this. 

Even in installation page it is showing 4.2.6 but 4.3.0 for snapshot.

mlvandijk:
@grasshopper7 The versions are specified in [data/versions.yaml](https://github.com/cucumber/docs.cucumber.io/blob/master/data/versions.yaml). Whenever there is a new release, they *should* be updated, but sometimes we forget (as docs and code are separate projects). I' ve just updated the versions in [this commit](https://github.com/cucumber/docs.cucumber.io/commit/43f28aec1b8d8ac0dbc79981138a66042790a745), so it should be ok now :)

grasshopper7:
@mlvandijk Version numbers are now updated.
I added the kotlin part. First time I tried anything in kotlin. Maybe there is a better approach.  Thanks.

bischoffdev:
This is a great start!
However, there are some points I want to mention (quickly read this document, I hope I will have more time later to do a more in-depth review) :-) 

* for the Maven dependencies it might be better to not specify concrete version numbers (in the document you use 4.2.6 but 4.3.0 is already available)
* the thread IDs are not fixed (not always 13 and 14 like in the example) so this could be misleading
* there is actually another way to run the tests by having the configuration directly inside of a JUnit runner like so:
```
import cucumber.api.cli.Main;
import org.junit.Test;

import static org.hamcrest.CoreMatchers.is;
import static org.junit.Assert.assertThat;

public class RestTest_IT {

    @Test
    public void test() {
        String[] arguments = new String[]{
                "--threads", "4",
                "-p", "json:target/json/test.json",
                "-p", "timeline:target/cucumber-parallel-report",
                "-p", "usage:target/cucumber-usage.json",
                "-p", "pretty",
                "-g", "your.glue.package",
                "path/to/features/"
        };
        int exitStatus = Main.run(arguments,
                Thread.currentThread().getContextClassLoader());
        assertThat(exitStatus, is(0));
    }
}
```

mlvandijk:
@laxersaz FYI: your first comment regarding Maven versions should be fixed by using the hugo tags (as mentioned above). @grasshopper7 - you might want to merge master into your branch ;)

grasshopper7:
@laxersaz Added my comments inline to your observations. Thanks.

> This is a great start!
> However, there are some points I want to mention (quickly read this document, I hope I will have more time later to do a more in-depth review) :-)
> 
> * for the Maven dependencies it might be better to not specify concrete version numbers (in the document you use 4.2.6 but 4.3.0 is already available) -- **Fixed it.**

> * the thread IDs are not fixed (not always 13 and 14 like in the example) so this could be misleading -- **Think it is a Java thing regarding thread ids. Also it is mentioned, that the results would be similar rather than exact** 

> * there is actually another way to run the tests by having the configuration directly inside of a JUnit runner like so: -- **I think it belongs to a different guide dealing with cucumber-jvm execution methods. The run() method is called by the main() method (described in the CLI section), only the exit status is swallowed in the main() method. This can be used in a testng test method or in a java class also.**

grasshopper7:
> @laxersaz FYI: your first comment regarding Maven versions should be fixed by using the hugo tags (as mentioned above). @grasshopper7 - you might want to merge master into your branch ;)

@mlvandijk  My bad. Fixed it. Though the version needs to be updated again. Shows 4.3.0, though 4.3.1 is released.

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+title: Parallel Execution
+subtitle: Using multiple threads to reduce test times
+polyglot:
+ - java
+ - kotlin
+ 
+weight: 1800
+---
+
+Cucumber-JVM allows [parallel execution](https://cucumber.io/blog/announcing-cucumber-jvm-4-0-0/) across multiple threads since version 4.0.0. There are several options to incorporate this inbuilt feature in a Cucumber project.
```
mlvandijk:
```suggestion
Cucumber-JVM allows [parallel execution](https://cucumber.io/blog/announcing-cucumber-jvm-4-0-0/) across multiple threads since version 4.0.0. There are several options to incorporate this built-in feature in a Cucumber project.
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+
+- [JUnit](/docs/guides/parallel-execution/#junit)
+- [TestNG](/docs/guides/parallel-execution/#testng)
+- [CLI](/docs/guides/parallel-execution/#cli)
+
+For each of these options, this tutorial will look at the project setup, configuration settings, execution commands.
```
mlvandijk:
```suggestion
For each of these options, this tutorial will look at the project setup, configuration settings and execution commands.
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+
+
+# JUnit
+Cucumber can be executed in parallel using **JUnit and Maven test execution plugins**. In JUnit the **feature files are run in parallel rather than scenarios**, which means **all the scenarios in a feature file will be executed by the same thread**. One can use either Maven Surefire or Failsafe plugin for executing the runners.
```
mlvandijk:
```suggestion
Cucumber can be executed in parallel using **JUnit and Maven test execution plugins**. In JUnit the **feature files are run in parallel rather than scenarios**, which means **all the scenarios in a feature file will be executed by the same thread**. You can use either Maven Surefire or Failsafe plugin to execute the runners.
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+package parallel;
+
+import cucumber.api.java.BeforeStep;
+import cucumber.api.java.en.Given;
+
+public class Stepdefs {
+
+	@BeforeStep
+	public void beforeStep() throws InterruptedException {
+		Thread.sleep(1000);
```
mlvandijk: Why do we need this `sleep()` ?

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+</dependency>
+<dependency>
+	<groupId>io.cucumber</groupId>
+	<artifactId>cucumber-junit</artifactId>
+	<version>{{% version "cucumberjvm" %}}</version>
+	<scope>test</scope>
+</dependency>
+```
+
+- Create a **parallel** folder in `src/test/resources` path and add the two feature files (`scenarios.feature` and `scenariooutlines.feature`) inside it.
```
mlvandijk: Is the "parallel" folder needed (both in `src/test/resources` and `src/test/java`) for configuration reasons, or is this a nice to have / suggestion?

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+
+class Stepdefs : En {
+	
+	init {
+		BeforeStep() { ->
+			Thread.sleep(1000)
+		}
+		
+		Given("Step from {string} in {string} feature file") { scenario: String , file: String ->
+            println("Thread ID - %2d - %s from %s feature file.".format(Thread.currentThread().id,scenario,file))
```
mlvandijk: Optional: Can we use String templating here? Maybe not with formatting... See also: https://stackoverflow.com/questions/23086291/format-in-kotlin-string-templates (it's fine to keep this as is)

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+        }
+	}
+}
+```
+{{% /block %}}
+
+- Add a cucumber **runner** using the `RunWith` annotation in the `parallel` package in the `src/test/java` folder.
```
mlvandijk: same question here regarding configuration. Is `parallel` needed? (I.e. will it work if I don't put them in a directory / package named `parallel`)

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+	<artifactId>maven-surefire-plugin</artifactId>
+	<version>2.22.0</version>
+	<configuration>
+		<parallel>methods</parallel>
+		<useUnlimitedThreads>true</useUnlimitedThreads>
+	</configuration>
+</plugin>
+```
+
+- Use the Maven `install` or a suitable command to **execute** the `POM`. This should run in parallel threaded mode. You should see a result similar to below. It is important to note that **both the scenarios in the file (`scenarios.feature`) are executed by thread with ID 14**. Similarly **both the rows of the scenario outline in the file (`scenariooutlines.feature`) are executed by thread with ID 13**.
```
mlvandijk: Rename to `scenario-outlines.feature` for readability? (if so, do this consistently)

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+
+```shell
+Thread ID - 13 - Scenario Outline Row 1 from scenariooutlines feature file.
+Thread ID - 13 - Scenario Outline Row 2 from scenariooutlines feature file.
+Thread ID - 14 - Scenario 1 from scenarios feature file.
+Thread ID - 14 - Scenario 2 from scenarios feature file.
+```
+
+- To execute using a Maven **Failsafe plugin include the below configuration** in the `build` section to the `POM`. Rename the runner class to `RunCucumberIT`.  You can find further details [here](/docs/community/not-cucumber/#maven-execution-plugin).
```
mlvandijk, line 152: Ok, so if I understand correctly Surefire will run classes named *Test, and Failsafe will run classes named *IT?

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+The thread count in the above setting is **4 threads per core**. If you want this to be **4 threads across all cores** set the `perCoreThreadCount` to **false**.
+```shell
+<parallel>methods</parallel>
+<threadCount>4</threadCount>
+<perCoreThreadCount>false</perCoreThreadCount>
+```
+
+If you have **multiple runners** then you can set the parallel option to `classesAndMethods`, `methods` or `classes`. For a single runner the `classes` option would be similar to a sequential execution.
+
+For a **visual representation** you can add the **timeline report** with the plugin option to a `CucumberOptions` annotation in the runner class. Scroll to the end for an image of the report.
```
mlvandijk: I don't understand what you mean here? Could you be specific which option and how to add it?

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+```
+
+- To execute using a Maven **Failsafe plugin include the below configuration** in the `build` section to the `POM`. Rename the runner class to `RunCucumberIT`.  You can find further details [here](/docs/community/not-cucumber/#maven-execution-plugin).
+
+```shell
+<plugin>
+	<groupId>org.apache.maven.plugins</groupId>
+	<artifactId>maven-failsafe-plugin</artifactId>
+	<version>2.22.0</version>
+	<executions>
```
mlvandijk, line 163: I have added the `<executions>` section to our (Kotlin/Maven/JUnit) project and the build still passes. I'm not sure if and how I can see that the tests are run in parallel though?

mlvandijk: I noticed that you are explicitly printing the Threads in this guide; ofcourse I don't do that in my project :)

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+<threadCount>4</threadCount>
+<perCoreThreadCount>false</perCoreThreadCount>
+```
+
+If you have **multiple runners** then you can set the parallel option to `classesAndMethods`, `methods` or `classes`. For a single runner the `classes` option would be similar to a sequential execution.
+
+For a **visual representation** you can add the **timeline report** with the plugin option to a `CucumberOptions` annotation in the runner class. Scroll to the end for an image of the report.
+
+
+# TestNG
```
mlvandijk, line 205: There's some duplication in this section, compared to the previous section. I'm wondering if we could remove the duplication *without* making it confusing, but I'm not sure at this point...

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+	<parallel>classes</parallel>
+	<threadCount>4</threadCount>
+</configuration>
+```
+
+For a **visual representation** you can add the **timeline report** with the plugin option to a `CucumberOptions` annotation in the runner class. Scroll to the end for an image of the report.
+
+
+# CLI
+The `Main class` in the `cucumber.api.cli package` is used to execute the feature files. There is no need to create any runner class. The usage options for this class is mentioned [here](https://github.com/cucumber/cucumber-jvm/blob/v4.0.0/core/src/main/resources/cucumber/api/cli/USAGE.txt). The `--threads` option needs to be set to a value **greater than 1** to run in parallel. When the parallel mode is used, the scenarios and rows in a scenario outline will be run in multiple threads.
```
mlvandijk:
```suggestion
The `Main class` in the `cucumber.api.cli package` is used to execute the feature files. You can run this class directly from the command line; in that case, there is no need to create any runner class. The usage options for this class is mentioned [here](https://github.com/cucumber/cucumber-jvm/blob/v4.0.0/core/src/main/resources/cucumber/api/cli/USAGE.txt). The `--threads` option needs to be set to a value **greater than 1** to run in parallel. When the parallel mode is used, the scenarios and rows in a scenario outline will be run in multiple threads.
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+ 
+Below is the basic command to start the execution.
+```shell
+java -cp <classpath> cucumber.api.cli.Main -g <steps package> --threads <thread count> <path to feature files>
```
mlvandijk: With `<steps package>`, do you mean the package containing the step definitions? In that case I'd prefer to use `<step defs package>` instead.

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+```
+
+Follow the below steps to **execute the command from a terminal**.
```
mlvandijk:
```suggestion
Follow the steps below to **execute the command from a terminal**.
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+
+- Download the necessary cucumber jars from the cucumber.io Maven repository (links provided under version number) or, if available, copy from the local repository into a folder. These need to be provided in the **classpath**. The versions mentioned are compatible with the latest cucumber release 4.3.0.
```
mlvandijk: Please, let's **not** describe downloading the jars, as we really recommend not doing that. I realise you are doing this here to give an alternative to getting the dependencies through Maven, but doing so leads to many problems and also I'd rather not maintain this list in the documentation.
In fact, in that case it might be better to also leave out the sections about the Maven dependencies for JUnit and TestNG, and just link to the part of the docs where that is already described. Which will save some duplication also :)

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+	3.  `cucumber-java8.jar` [`4.3.0`](http://central.maven.org/maven2/io/cucumber/cucumber-java8/4.3.0/cucumber-java8-4.3.0.jar)
+	4.  `cucumber-html.jar` [`0.2.7`](http://central.maven.org/maven2/io/cucumber/cucumber-html/0.2.7/cucumber-html-0.2.7.jar)
+	5.  `gherkin.jar` [`5.1.0`](http://central.maven.org/maven2/io/cucumber/gherkin/5.1.0/gherkin-5.1.0.jar)
+	6.  `cucumber-expressions.jar` [`6.2.2`](http://central.maven.org/maven2/io/cucumber/cucumber-expressions/6.6.2/cucumber-expressions-6.6.2.jar)
+	7.  `datatable.jar` [`1.1.12`](http://central.maven.org/maven2/io/cucumber/datatable/1.1.12/datatable-1.1.12.jar)
+	8.  `datatable-dependencies.jar` [`1.1.12`](http://central.maven.org/maven2/io/cucumber/datatable-dependencies/1.1.12/datatable-dependencies-1.1.12.jar)
+	9.  `tag-expressions.jar` [`1.1.1`](http://central.maven.org/maven2/io/cucumber/tag-expressions/1.1.1/tag-expressions-1.1.1.jar)
+	10. `typetools.jar` [`0.5.0`](http://central.maven.org/maven2/net/jodah/typetools/0.5.0/typetools-0.5.0.jar)
+	
+- Create a folder named `cukecli` and a folder `parallel` inside it, to keep the step definition classes and feature files.
```
mlvandijk: I'd prefer a `cucumber` or `cucumbercli` directory (personal preference, I guess)

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+
+- Create the **two feature files** (`scenarios.feature` and `scenariooutlines.feature`) inside the `parallel` folder as described in the JUnit section.
+
+- Create the **step definition class** in the `parallel` folder as described in the JUnit section.
+
+- Open up a **terminal window** and navigate to the source folder of the project, in our case `cukecli`.
+
+{{% block "java" %}}
```
mlvandijk, line 315: When the content of the language blocks is the same, you can specify multiple languages in the same block, separated by a comma, like this `{{% text "java,kotlin" %}}` (see [Contributing to the documentation](https://cucumber.io/docs/community/contributing-to-documentation/#paragraphs-or-text-that-is-valid-for-multiple-programming-languages))

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,396 @@
...
+- Compile the step definition class. Add **path to each of the downloaded cucumber jars to the classpath** using the **-cp** option.
+
+```shell
+kotlinc -cp .;<path to each cucumber jar> -jvm-target 1.8 ./parallel/Stepdefs.java
+```
+{{% /block %}}
+
+- Execute using the below command.
+
+{{% block "java" %}}
```
mlvandijk, line 335: Same here.
Note that anything **outside** the language blocks would be visible on all polyglot pages (if we add javascript/ruby in the future).

mlvandijk (changes requested):
Hi @grasshopper7 - this is a more elaborate addition than I expected; thank you! Sorry it took my longer than I would've liked to look into it in more detail. Overall it's really great. I do have some minor textual changes, and am wondering if we could make sure to limit duplication (i.e. not say the same thing in multiple places, but add links).

grasshopper7:
> Hi @grasshopper7 - this is a more elaborate addition than I expected; thank you! Sorry it took my longer than I would've liked to look into it in more detail. Overall it's really great. I do have some minor textual changes, and am wondering if we could make sure to limit duplication (i.e. not say the same thing in multiple places, but add links).

@mlvandijk I committed the minor text changes. Will have a look into the others later today.

luke-hill:
Just for continued / future reference @grasshopper7  (As this has gotten to be a heavily committed change), you're able to push up multiple commits in one go to the repo (Say 5 local commits in one go with a `git push`). Just so people monitoring don't need to check back as often

mlvandijk:
@luke-hill The multiple commits are probably due to @grasshopper7 accepting my suggested changes..

content/docs/guides/parallel-execution.md
grasshopper7: Ooops... I had added this to slow things down. Forgot to remove it. No need at all. Will remove it.

content/docs/guides/parallel-execution.md
grasshopper7: No configuration reason. Just a package name I thought would be appropriate. Will mention that folder can be named anything.

content/docs/guides/parallel-execution.md
grasshopper7: No configuration reason. Just a package name I thought would be appropriate. Will mention that folder can be named anything.

content/docs/guides/parallel-execution.md
grasshopper7: Yeah. The plugins pick up these class names matching the pattern automatically. There are a few other patterns mentioned in the linked article

content/docs/guides/parallel-execution.md
grasshopper7: Thread id was the easiest I could think off. You can configure the timeline plugin to get a chart. But that would be bit off an overkill.

content/docs/guides/parallel-execution.md
grasshopper7: Maybe a separate section in this tutorial which mentions how to set this up and what is the result. Then can be linked from the other parts.

content/docs/guides/parallel-execution.md
grasshopper7: How about the glue code package?

content/docs/guides/parallel-execution.md
grasshopper7: Will change it.

content/docs/guides/parallel-execution.md
grasshopper7: Will change all occurrences.

grasshopper7:
> Just for continued / future reference @grasshopper7 (As this has gotten to be a heavily committed change), you're able to push up multiple commits in one go to the repo (Say 5 local commits in one go with a `git push`). Just so people monitoring don't need to check back as often

@luke-hill I accepted the changes suggested. Is there a way to avoid multiple commits for this? Thx.

content/docs/guides/parallel-execution.md
grasshopper7: Removed the dependencies in CLI section.
For the other two the 10 minute tutorial uses the cucumber-archetype. Is that updated to get latest versions? I tried it out around 3 months back and it added some pretty old version,

content/docs/guides/parallel-execution.md
grasshopper7: Will change this

content/docs/guides/parallel-execution.md
grasshopper7: Will change this

grasshopper7:
@mlvandijk Made the changes other than the string templating and multiple runners.

content/docs/guides/parallel-execution.md
mlvandijk: The archetype has been updated to a more recent version (v4.?) but probably not *the* most recent version... It's a great thing to use to get started, but a pain to maintain.

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,389 @@
...
+{{% /block %}}
+
+# JUnit
+
+{{% block "java,kotlin" %}}
+Cucumber can be executed in parallel using **JUnit and Maven test execution plugins**. In JUnit the **feature files are run in parallel rather than scenarios**, which means **all the scenarios in a feature file will be executed by the same thread**. You can use either Maven Surefire or Failsafe plugin to execute the runners.
+
+- Create a Maven project in your favorite IDE. Add `cucumber-java8` and `cucumber-junit` **dependencies** to the `POM`.
+
+```shell
```
mlvandijk: We also mention this [here](https://cucumber.io/docs/installation/java/#maven) and [here](https://cucumber.io/docs/tools/java/#maven) - I guess it's ok to duplicate though.

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,389 @@
...
+<dependency>
+	<groupId>io.cucumber</groupId>
+	<artifactId>cucumber-java8</artifactId>
+	<version>{{% version "cucumberjvm" %}}</version>
+	<scope>test</scope>
+</dependency>
+<dependency>
```
mlvandijk: We also mention this [here](https://cucumber.io/docs/cucumber/checking-assertions/#junit). Again, I think it's ok to duplicate, otherwise the reader would be clicking back and forth.

Note that the JUnit dependency itself is missing on this page - could you add it here too?

content/docs/guides/parallel-execution.md
mlvandijk: Ok, thnx :)

content/docs/guides/parallel-execution.md
mlvandijk: Ok, thanks. That's what I thought, I just wanted to make sure. If it's not needed, maybe don't mention creating a folder at all, to prevent confusion?

mlvandijk:
@grasshopper7 Could you merge master into this branch again (I think the versions have been updated again). I'll try to create a parallel project from this guide to make sure I get actual parallel execution and have a look to see if I can think of a way to prevent duplication on the page (to keep it easy to maintain in the future).

content/docs/guides/parallel-execution.md
grasshopper7: The class files are kept in the parallel package. If the feature files are kept in the parallel folder these are picked up automatically without adding any configuration to the CucumberOptions annotation.

content/docs/guides/parallel-execution.md
grasshopper7: Done will add to commit

content/docs/guides/parallel-execution.md
grasshopper7: Also the execution times should decrease.

content/docs/guides/parallel-execution.md
grasshopper7: Using the archetype is a good solution. As long as it gets a V4 it works, unless in a future version parallel functionality is removed. Will add the link.

content/docs/guides/parallel-execution.md
grasshopper7: Will add this and the archetype option. Reader can pick anyone of them.

content/docs/guides/parallel-execution.md
grasshopper7: No need to add the junit or testng dependency explicitly in pom. The cucumber-junit or cucumber-testng pulls them as a transitive one.

grasshopper7: Though the cucumber-junit dependency is missing at this [link](https://cucumber.io/docs/installation/java/#maven).

grasshopper7:
@mlvandijk Removed the dependency addition for  JUnit and testng. Linked to existing documentatio. Updated kotlin string template. Removed duplication of timeline reporter to separate section. Tried to simplify multiple runners.

grasshopper7:
@mlvandijk Removed the dependency addition for  JUnit and testng. Linked to existing documentatio. Updated kotlin string template. Removed duplication of timeline reporter to separate section. Tried to simplify multiple runners. 

Clicked the close button by mistake. Reopened.

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+title: Parallel Execution
+subtitle: Using multiple threads to reduce test times
+polyglot:
+ - java
+ - kotlin
+ 
+weight: 1800
+---
+{{% block "java,kotlin" %}}
+Cucumber-JVM allows [parallel execution](https://cucumber.io/blog/announcing-cucumber-jvm-4-0-0/) across multiple threads since version 4.0.0. There are several options to incorporate this built-in feature in a Cucumber project.
```
mlvandijk:
```suggestion
Cucumber-JVM allows parallel execution across multiple threads since [version 4.0.0.](https://cucumber.io/blog/announcing-cucumber-jvm-4-0-0/) There are several options to incorporate this built-in feature in a Cucumber project. You can do so using:
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+
+For each of these options, this tutorial will look at the project setup, configuration settings and execution commands.
+{{% /block %}}
+
+# JUnit
+
+{{% block "java,kotlin" %}}
+Cucumber can be executed in parallel using **JUnit and Maven test execution plugins**. In JUnit the **feature files are run in parallel rather than scenarios**, which means **all the scenarios in a feature file will be executed by the same thread**. You can use either Maven Surefire or Failsafe plugin to execute the runners.
+
+- Create a Maven project in your favorite IDE using the [cucumber-archetype](/docs/guides/10-minute-tutorial/#create-an-empty-cucumber-project) or adding Cucumber dependencies to the POM as detailed [here](https://cucumber.io/docs/installation/java/#maven) and Junit dependencies [here](/docs/cucumber/checking-assertions/#junit).
```
mlvandijk:
```suggestion
- Create a Maven project in your favorite IDE using the [cucumber-archetype](/docs/guides/10-minute-tutorial/#create-an-empty-cucumber-project) or by adding Cucumber dependencies to the POM as detailed [here](https://cucumber.io/docs/installation/java/#maven) and Junit dependencies [here](/docs/cucumber/checking-assertions/#junit).
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+package parallel;
+
+import cucumber.api.java.BeforeStep;
+import cucumber.api.java.en.Given;
+
+public class Stepdefs {
+
+	@Given("Step from {string} in {string} feature file")
+	public void step(String scenario, String file) {
+		System.out.format("Thread ID - %2d - %s from %s feature file.\n", Thread.currentThread().getId(), scenario,file);
```
mlvandijk:
```suggestion
		System.out.format("Thread ID - %2d - %s from %s feature file.\n",
		Thread.currentThread().getId(), scenario,file);
```

mlvandijk: Prevent (or at least limit) horizontal scrolling.

content/docs/guides/parallel-execution.md
mlvandijk: Much better now, thanks!

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+	<parallel>classes</parallel>
+	<threadCount>4</threadCount>
+</configuration>
+```
+{{% /block %}}
+
+# CLI
+
+{{% block "java,kotlin" %}}
+The `Main class` in the `cucumber.api.cli package` is used to execute the feature files. You can run this class directly from the command line; in that case, there is no need to create any runner class. The usage options for this class is mentioned [here](https://github.com/cucumber/cucumber-jvm/blob/v4.0.0/core/src/main/resources/cucumber/api/cli/USAGE.txt). The `--threads` option needs to be set to a value **greater than 1** to run in parallel. When the parallel mode is used, the scenarios and rows in a scenario outline will be run in multiple threads.
```
mlvandijk:
```suggestion
The `Main class` in the `cucumber.api.cli package` is used to execute the feature files. You can run this class directly from the command line; in that case, there is no need to create any runner class. The usage options for this class are mentioned [here](https://github.com/cucumber/cucumber-jvm/blob/v4.0.0/core/src/main/resources/cucumber/api/cli/USAGE.txt). The `--threads` option needs to be set to a value **greater than 1** to run in parallel. When the parallel mode is used, the scenarios and rows in a scenario outline will be run in multiple threads.
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+ 
+Below is the basic command to start the execution.
+```shell
+java -cp <classpath> cucumber.api.cli.Main -g <glue package> --threads <thread count> <path to feature files>
```
mlvandijk, line 303: Could you please **also** provide an example of what this command would like look like with the setup as described in this guide (i.e. using directory `parallel`). I'm not too familiar with running Cucumber from the command line so I cannot get this to work.

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+```
+
+Follow the steps below to **execute the command from a terminal**.
+
+- Download the cucumber jars from the cucumber.io repository into a folder. These need to be provided in the **classpath**.
```
mlvandijk: This really needs to be removed. We don't recommend downloading individual jars, as it's very easy to get the wrong dependencies which leads to problems/questions.

mlvandijk: In my opinion, we only need to mention how to **run** in parallel from the CLI, using the `--threads <number of threads>` option.

mlvandijk:
```suggestion
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+	
+- Create a folder named `cucumbercli` and a folder `parallel` (or any folder names) inside it, to keep the step definition classes and feature files.
```
mlvandijk:
```suggestion
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+
+- Add the two feature files (`scenarios.feature` and `scenario-outlines.feature`) and **step definition class** as described in the JUnit section.
+
+- Open up a **terminal window** and navigate to the source folder of the project, in our case `cucumbercli`.
```
mlvandijk:
```suggestion
- Open up a **terminal window** and navigate to the source folder of the project.
```

mlvandijk (commented):
Finally got around to check this guide step by step. Here is my progress so far:
[x] Java - Junit - Maven
[x] Java - TestNg - Maven
[ ] Java - Cli
I'm not too familiar with running from the CLI so I couldn't get that to work yet (see questions below).
If you could clarify, and/or someone else can confirm that it works, that would be great.

**Update:** also works with Kotlin, as long as StepDefs are in `src/test/**java**` (maybe it's possible to configure Surefire and/or Failsafe to use  `src/test/**kotlin**`, but I don't know how)
[x] Kotlin - Junit - Maven
[x] Kotlin - TestNg - Maven
[ ] Kotlin - Cli

Comments:
* Mainly minor textual edits; please accept if you agree.
* Question about running from the CLI
* The part about the jars really needs to be removed. (Could you mention where to find the jars if they're downloaded with Maven?
* A few edits for the Kotlin part; please accept.

**Suggestion:** Please consider splitting this PR and moving the CLI part to a separate PR, so we can merge the JUnit/TestNG parts (with Kotlin edits), which imho are a great addition!

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+{{% block "kotlin" %}}
+
+```kotlin
+package parallel
+
+import cucumber.api.java8.En
+
+class Stepdefs : En {
+		
+		Given("Step from {string} in {string} feature file") { scenario: String , file: String ->
```
mlvandijk:
```suggestion
		    init {
        Given("Step from {string} in {string} feature file") { scenario: String , file: String ->
            println("Thread ID - ${Thread.currentThread().id} - $scenario from $file feature file")
        }
    }
```

mlvandijk: When using lambda's with Kotlin, the stepdefs need to be inside an `init {}` block.
We could consider adding examples both with annotations and lambdas, but that could be a separate additional PR; let's get this one merged asap. :)

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+
+{{% block "java" %}}
+
+```java
+package parallel;
+
+import cucumber.api.java.BeforeStep;
+import cucumber.api.java.en.Given;
+
+public class Stepdefs {
```
mlvandijk:
```suggestion
public class StepDefs {
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+{{% /block %}}
+
+{{% block "kotlin" %}}
+
+```kotlin
+package parallel
+
+import cucumber.api.java8.En
+
+class Stepdefs : En {
```
mlvandijk:
```suggestion
class StepDefs : En {
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+- Add the two feature files (`scenarios.feature` and `scenario-outlines.feature`) and **step definition class** as described in the JUnit section.
+
+- Open up a **terminal window** and navigate to the source folder of the project, in our case `cucumbercli`.
+{{% /block %}}
+
+{{% block "java" %}}
+- Compile the step definition class. Add the **path to the folder containing cucumber jars to the classpath** using the **-cp** option.
+
+```shell
+javac -cp .;<path to cucumber jar folder>/* ./parallel/Stepdefs.java
```
mlvandijk:
```suggestion
javac -cp .;<path to cucumber jar folder>/* ./parallel/StepDefs.java
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+```
+{{% /block %}}
+
+{{% block "kotlin" %}}
+- Compile the step definition class. Add **path to each of the downloaded cucumber jars to the classpath** using the **-cp** option.
+
+```shell
+kotlinc -cp .;<path to each cucumber jar> -jvm-target 1.8 ./parallel/Stepdefs.java
```
mlvandijk:
```suggestion
kotlinc -cp .;<path to each cucumber jar> -jvm-target 1.8 ./parallel/StepDefs.kt
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+{{% block "kotlin" %}}
+
+```kotlin
+package parallel
+
+import cucumber.api.junit.Cucumber
+import org.junit.runner.RunWith
+
+@RunWith(Cucumber::class)
+class RunCucumberTest {
```
mlvandijk:
```suggestion
class RunCucumberTest
```

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+}
```
mlvandijk:
```suggestion

```

mlvandijk: These brackets are optional in Kotlin for empty class.

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+	}
+}
+```
+{{% /block %}}
+
+{{% block "java,kotlin" %}}
+- Add a cucumber **runner** using the `RunWith` annotation in the `parallel` package (same name as step definition package) in the `src/test/java` folder.
+{{% /block %}}
+
+{{% block "java" %}}
```
mlvandijk, line 97: Afaik you don't need both the {{% block "java" %}} and the ```java around the code for it to only show up in the right language tab, but do try that out first..

mlvandijk (commented):
Works with Kotlin, but does need a few edits in the Kotlin code / sections.

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+	<groupId>org.apache.maven.plugins</groupId>
+	<artifactId>maven-surefire-plugin</artifactId>
+	<version>2.22.0</version>
+	<configuration>
+		<parallel>methods</parallel>
+		<useUnlimitedThreads>true</useUnlimitedThreads>
+	</configuration>
+</plugin>
+```
+
```
mlvandijk, line 142: {{% block "kotlin" %}}
For Surefire to find your step definitions, make sure they are in `src/test/**java**`.
{{% /block %}}

mlvandijk: Alternatively, it might be possible to configure Surefire to use `src/test/kotlin`, but I don't know how.

content/docs/guides/parallel-execution.md
```diff
@@ -0,0 +1,380 @@
...
+```shell
+Thread ID - 13 - Scenario Outline Row 1 from scenario-outlines feature file.
+Thread ID - 13 - Scenario Outline Row 2 from scenario-outlines feature file.
+Thread ID - 14 - Scenario 1 from scenarios feature file.
+Thread ID - 14 - Scenario 2 from scenarios feature file.
+```
+
+- To execute using a Maven **Failsafe plugin include the below configuration** in the `build` section to the `POM`. Rename the runner class to `RunCucumberIT`.  You can find further details [here](/docs/community/not-cucumber/#maven-execution-plugin).
+
```
mlvandijk, line 153:
```suggestion
{{% block "kotlin" %}}
For Failsafe to find your step definitions, make sure they are in src/test/**java**.
{{% /block %}}
```

mlvandijk: Alternatively, it might be possible to configure Failsafe to use `src/test/kotlin`, but I don't know how.

mlvandijk (commented):
Additional Kotlin edits.

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Added to batch commit

content/docs/guides/parallel-execution.md
grasshopper7: Removed.

mlvandijk (approved):
Thanks so much for all your work on this @grasshopper7 - really appreciate it!!

grasshopper7:
@mlvandijk I found some issues in this. Just fixing them. Will create a new PR.

mlvandijk:
Sure, thanks!

PR #348, opened 2019-05-15
Add Gherkin formatter/validator to the list

glibas:


luke-hill (commented):
The example given on that website looks a bit wrong and incorrect. Seems a bit odd to be recommending a site to check linting e.t.c. with an example that is non-comformant?

glibas:
Hi @luke-hill , 

Fair point, thank you. Example fixed. Originally the idea was to have some errors to highlight linting and formatting functionality. So that users can try out hotkeys straight away.

glibas:
Hi @luke-hill could you please take a look one more time

luke-hill:
Some of it is still a bit iffy, but for now it seems ok. I'll merge this in.

PR #357, opened 2019-07-11
Update to cucumber 4.5

beardedqa:
As after cucumber some imports have been changed, change in documentation is required. 

**Please do not merge this change** before the IntelliJ cucumber java plugin update https://youtrack.jetbrains.com/issue/IDEA-217391


luke-hill:
@beardedqa - Are you happy for this to be merged in now?

I'd rather a java person give it the once over as well, but you mentioned about waiting until the plugin update. The thread seems to elude to the fact this is done.

data/versions.yaml
```diff
@@ -4,8 +4,8 @@
 #
 
 # Note: When updating cucumber-jvm version, please update both the snapshot version and the regular version
-cucumberjvm-snapshot: "4.4.0-SNAPSHOT"
-cucumberjvm: "4.3.1"
+cucumberjvm-snapshot: "4.5.5-SNAPSHOT"
```
mlvandijk: I'd like to at least merge the version updates (if not the import statements just yet). Would you mind separating these changes into 2 different PRs? (We could update import statements when v5 hits? And we'd need to make sure to update consistently all throughout the docs)

mlvandijk (changes requested):

data/versions.yaml
mpkorstanje: Better make a new PR then. We're at 4.7.1 right now.

mpkorstanje: With 5.0.0-SNAPSHOT as the next planned release.

mlvandijk: Good to know, thnx :)

luke-hill:
@beardedqa can you drop the 2nd commit and then rebase the 1 remaining commit.

mlvandijk:
@beardedqa Would you mind removing the version change from this PR? We could then merge the other changes. (Would love help checking the rest of the docs for changes needed btw)

mpkorstanje:
I've merged it for now. With v5.0.0-RC1 coming up I'd rather not wait any longer for IDEA to catch up.

Cheers!

PR #359, opened 2019-07-11
SEC: theme node packages mem, lodash.mergewith updated

olleolleolle:
This PR adds[ **`package.json` resolutions**](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/), to avoid security warnings about issues in the Node packages for the Theme we are using.

Easy fix. Benefit: avoid warnings about security issues reported in #358

### DIFF
--- a/themes/cucumber-hugo/package.json
+++ b/themes/cucumber-hugo/package.json
@@ -11,6 +11,10 @@
     "postcss-cli": "^5.0.0",
     "rimraf": "^2.6.2"
   },
+  "resolutions": {
+    "mem": "^4.0.0",
+    "lodash.mergewith": "^4.6.2"
+  },
   "scripts": {
     "build": "yarn build-clean && yarn build-sass && yarn build-autoprefix",
     "build-clean": "rimraf static/css/cucumber.css*",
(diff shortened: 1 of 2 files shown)


mlvandijk:
@aslakhellesoy Could you have a look at this, please? I'm not familiar with these particular tools.

mxygem:
I'm all for fixing security issues. @olleolleolle Have you tried building the site locally to see if anything breaks? Netlify doesn't build forks afaik, so that prevents me from easily checking it.

luke-hill:
My philosophy on these kinda changes is if they pass CI and doesn't break anything merge it in. We have a revert ability.

aslakhellesoy:
Well said @luke-hill.

@olleolleolle the `resolutions` feature is only supported by yarn (not by npm). People using npm will still see warnings.

olleolleolle:
@aslakhellesoy I'm not aware of any equivalent to "second-order dependency hints" (my description of what `resolutions` does) for npm. Is there one?

olleolleolle:
Ah, and about warnings displayed: this repo has `yarn.lock` (which GitHub can inspect) but no `package-lock.json`.

(I was mostly referring to the big yellow banner on the GitHub repository, about security warnings.)

mxygem (approved):
Thanks for helping with this. :) If things break, we can always revert.

PR #365, opened 2019-07-30
organisation -> organization

rlin2k1:


aslakhellesoy:
I don't think we've made a decision to switch from UK English to US English yet.

PR #366, opened 2019-08-01
Update step-definitions.md

AReyes-cl:
Fix java regexp notation in expression

### DIFF
--- a/content/docs/cucumber/step-definitions.md
+++ b/content/docs/cucumber/step-definitions.md
@@ -118,7 +118,7 @@ A step definition's *expression* can either be a [Regular Expression](https://en
 
 {{% block "java" %}}
 ```java
-@Given("I have {int} cukes in my belly")
+@Given("I have (d+) cukes in my belly")
 public void i_have_n_cukes_in_my_belly(int cukes) {
 }
 ```


olleolleolle (commented):
If you read the paragraph after the code samples, they seem to be about Cucumber Expressions, which are surrounded by curly braces. The sentence about capture group handling explains how to use that with the {int} expression. 

It seems to me that these code samples should be Cucumber Expressions all.

luke-hill:
Yeh the four capture blocks have 2 as regexp and 2 as cukeexp.

Might I suggest a change that adds both options to all four blocks?

mlvandijk:
Reading the text surrounding the example as well as the other examples, they are intended to use Cucumber expressions, and not regular expressions. 
I do understand the text around the example could be a bit confusing as it mentions regex just before the example and then goes on to explain how `int` is a built in type that translates to `\d+`. 
Perhaps it would be better to put all text above the example, to make it easier to understand?

mlvandijk (changes requested):
Example should use Cucumber expressions, not regex.
Instead of changing the example, let's rearrange the text surrounding the example (see my comment).

mlvandijk:
@AReyes-cl Are you able to pick up the suggested changes?

mlvandijk:
Closing this PR in favor of #430

PR #370, opened 2019-08-16
Add Java, Kotlin examples on how to produce code snippets for missing step definitions

ralphkar:
The latest Cucumber requires to use the summary-plugin for printing code snippets for missing step definitions. I've changed the documentation so that it is documented.


content/docs/cucumber/api.md
```diff
@@ -1041,6 +1041,35 @@ class RunCucumberTest {
 ```
 The default option for `strict` is `false`.
 
+For example if you want to tell Cucumber to print code snippets for missing
+step definitions use the plugin `summary`, you can specify it like this:
```
luke-hill: step definitions, you can use the `summary` plugin. It is specified like this ...

luke-hill (changes requested):

ralphkar:
The PR is now hanging here for almost 2 weeks. Should I just merge it???

content/docs/cucumber/api.md
```diff
@@ -1041,6 +1041,35 @@ class RunCucumberTest {
 ```
 The default option for `strict` is `false`.
 
+For example if you want to tell Cucumber to print code snippets for missing
+step definitions use the `summary` plugin, you can specify it like this:
+
+```java
```
olleolleolle, line 1048: This example comes in two programming language variants.

If you surround each of the examples with a language block:

```
{{% block "java" %}}
Place the java example here, backticks and all.
{{% /block %}}
```

and 


```
{{% block "kotlin" %}}
Place the kotlin example here, backticks and all.
{{% /block %}}
```

This would make the documentation more usable.

ralphkar: As far as I see it, there is already java and kotlin block (quoted with three backticks). When running hugo only one is displayed (even though both are identical). So what is the difference between the three backticks and using the {{% block ...}} directive?

olleolleolle: One is code syntax highlighting, one is for creating UI elements that indicate to the reader which language is used.

ralphkar: I see. Thx for the explanation. I made the change.

olleolleolle (approved):

PR #374, opened 2019-08-24
As discussed in PR "Update to cucumber 4.5" #357

nhojpatrick:


luke-hill (approved):
Sure? I guess. I'll let a JVM maintainer decide if these numbers are right.

I'm guessing they point to a stable version?

mpkorstanje:
The numbers are right!

PR #396, opened 2019-10-13
clarify empty behaviour in string expressions

davidjgoss:
Goes with cucumber/cucumber#754, where we are making empty string behaviour consistent across implementations.


luke-hill (approved):

luke-hill:
Good thing to add this is. I've used this behaviour briefly alongside anonymous captures.

The other PR can be thought of as a bugfix, so this is good to merge in as is.

PR #398, opened 2019-10-17
Update with api changes for v5.0.0

mpkorstanje:


data/versions.yaml
```diff
@@ -3,10 +3,8 @@
 # They can be referred to with the custom Hugo shortcode, e.g. {{% version "cucumberjvm" %}}
 #
 
-# Note: When updating cucumber-jvm version, please update both the snapshot version and the regular version
-cucumberjvm-snapshot: "5.0.0-SNAPSHOT"
-cucumberjvm: "4.7.4"
-cucumberarchetype: "4.2.6.1"
+cucumberjvm: "5.0.0-RC1"
```
luke-hill: should we be documenting / referencing this? We don't for ruby (Use the rc1 package)

mpkorstanje: Good point. I'll merge this branch when v5.0.0 is released and change that.

mpkorstanje (commented):
Stuff that bothers me:

- [x] We are using `cucumber-java` and `cucumber-java8` inconsistently for the Kotlin examples.
- [ ] DocString and DataTableType need their own section on the same level as parameter types (configuration is the wrong place).
- [ ] ParameterType section for Java and Kotlin needs to explain they should be on the glue path. No need to reference configuration
 - [x] Java/Kotlin ParameterType annotations do not require all params by default, but don't show them currently. 
- [x] We use `com.example.app` and `com.example` inconsistently.

mpkorstanje:
>  DocString and DataTableType need their own section on the same level as parameter types (configuration is the wrong place).
>  ParameterType section for Java and Kotlin needs to explain they should be on the glue path. No need to reference configuration

It is impossible to explain data table and doc transformers withouth also explaing what they are. Fixing this would require splitting Cucumber Reference and turning Steps, DataTables, DocString, Hooks, Tags and Running Cucumber into top level sections.

I'd rather leave that out of this PR.

olleolleolle:
💌  This has become awesome and tiny, **just API differences**! Thanks for narrowing this down using that other PR. 💌

content/docs/cucumber/configuration.md
```diff
@@ -15,170 +15,376 @@ polyglot:
 # Type Registry
 
 {{% block "java,kotlin" %}}
-The type registry is used to configure parameter types and data table types. It can be configured by placing an implementation
-of `cucumber.api.TypeRegistryConfigurer` on the glue path.
+Parameter types let you convert parameters from cucumber-expressions to objects.
+Data table and doc string string types let you convert data tables and doc
```
mlvandijk: Repeat of "string"; is this intended or not? If so, maybe put `doc string` in back tics to indicate it is terminology.

content/docs/cucumber/configuration.md
```diff
@@ -15,170 +15,376 @@ polyglot:
...
+strings to objects. Like step definitions, type definitions are part of the glue.
+When placed on the glue path Cucumber will detect them automatically.
```
mlvandijk, line 21: Do we explain "glue" and "glue path" anywhere? Am thinking a glossary might be helpful (probably outside of the scope of this PR)

content/docs/cucumber/configuration.md
```diff
@@ -15,170 +15,376 @@ polyglot:
...
-The type registry is used to configure parameter types and data table types. It can be configured by placing an implementation
-of `cucumber.api.TypeRegistryConfigurer` on the glue path.
+Parameter types let you convert parameters from cucumber-expressions to objects.
+Data table and doc string string types let you convert data tables and doc
+strings to objects. Like step definitions, type definitions are part of the glue.
+When placed on the glue path Cucumber will detect them automatically. 
+{{% /block %}}
+
+{{% block "java" %}}
+For example, the following class registers custom "Author" data table type:
```
mlvandijk:
```suggestion
For example, the following class registers a custom "Author" data table type:
```

content/docs/cucumber/configuration.md
```diff
@@ -15,170 +15,376 @@ polyglot:
...
+	        objectMapper.readValue(docString, JsonNode::class)
+	    }
     }
 }
 ```
+{{% /block %}}
+
+{{% block "java,kotlin" %}}
+Using the `@DefaultParameterTransformer`, `@DefaultDataTableEntryTransformer`
+and `DefaultDataTableCellTransformer` annotations also possible to plugin an
```
mlvandijk:
```suggestion
and `@DefaultDataTableCellTransformer` annotations, it is also possible to plug in an 
```

content/docs/cucumber/configuration.md
```diff
@@ -15,170 +15,376 @@ polyglot:
...
+    fun transformer(fromValue: Any, toValueType: Type): Any {
+        return objectMapper.convertValue(fromValue, objectMapper.constructType(toValueType))
     }
+}
+```
+{{% /block %}}
+
+{{% block "java,kotlin" %}}
+For lambda defined step definitions, there are `DefaultParameterTransformer`, 
+`DefaultDataTableCellTransformer` and `DefaultDataTableEntryTransformer` functions:
```
mlvandijk:
```suggestion
`DefaultDataTableCellTransformer` and `DefaultDataTableEntryTransformer` {{% stepdef-body %}}:
```

mlvandijk: Use this short code to correctly display "method" or "function" for the respective programming language. (Although I'm not sure if/how that would work with a text for 2 different languages)

content/docs/cucumber/configuration.md
```diff
@@ -208,7 +414,7 @@ end
 {{% /block %}}
 
 {{% block "java,kotlin,ruby" %}}
-If you are using a type that has not yet been defined, you will get an error similar to:
+If you are using a type that has not yet been defined, you will got an error similar to:
```
mlvandijk:
```suggestion
If you are using a type that has not yet been defined, you will get an error similar to:
```

mlvandijk: Accidental change?

content/docs/cucumber/cucumber-expressions.md
```diff
@@ -82,23 +82,19 @@ we can define a custom parameter type in Cucumber's [configuration](/docs/cucumb
 
 {{% block "java" %}}
 ```java
-typeRegistry.defineParameterType(new ParameterType<>(
-    "color",           // name
-    "red|blue|yellow", // regexp
-    Color.class,       // type
-    Color::new         // transformer function
-))
+@ParameterType("red|blue|yellow")  // regexp
```
mlvandijk, line 85: Oh, I didn't know I could do this. This looks very nice & useful!

data/versions.yaml
mlvandijk: Maybe put WIP to prevent accidental merge?

mlvandijk (changes requested):

content/docs/cucumber/cucumber-expressions.md
mpkorstanje: In V5 you can.

mpkorstanje:
I think this will do for now. I really would like to rework the Cucumber API section but I'd rather not combine it with these changes.

PR #407, opened 2019-10-24
Remove BDD Tutorial

mpkorstanje:
The content of the BDD tutorial seems to be mostly duplicated by the
BDD section of the documentation. However the BDD section recommends
example mapping while the BDD tutorial talks about discovery workshops.


mattwynne:
Thanks @mpkorstanje - I hadn't spotted this page yet.

PR #412, opened 2019-10-31
Fix Gherkin language execution parameters

brazilianbytes:
Cucumber 6.0.3 parameters are different from documentation.

### DIFF
--- a/content/docs/gherkin/reference.md
+++ b/content/docs/gherkin/reference.md
@@ -442,21 +442,28 @@ In order to allow Gherkin to be written in a number of languages, the keywords h
 
 ### List translation options
 You can get information about the translations from the command line.
+Note: 'cucumber' command may be different over code languages. ex. for Node.js is 'cucumber-js'
 
 To see a listing of available languages:
 ```shell
-cucumber --i18n help
+cucumber --i18n-languages help
 ```
 
 To list the keywords of a particular language, use the language code:
 ```shell
-cucumber --i18n <language_code>
+cucumber --i18n-keywords <language_code>
 ```
 
 For example, to see the keywords in French:
 ```shell
-cucumber --i18n fr
+cucumber --i18n-keywords fr
 ```
+
+Finally, to run the test using a particular language, use  'language' parameter and language code:
+```shell
+cucumber --language fr
+```
+
 
 ### Overview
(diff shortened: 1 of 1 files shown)


content/docs/gherkin/reference.md
```diff
@@ -442,22 +442,29 @@ In order to allow Gherkin to be written in a number of languages, the keywords h
 
 ### List translation options
 You can get information about the translations from the command line.
+Note: 'cucumber' command may be different over code languages. ex. for Node.js is 'cucumber-js'
 
 To see a listing of available languages:
 ```shell
-cucumber --i18n help
+cucumber --i18n-languages help
```
aslakhellesoy, line 449: Unfortunately, this option is named differently across platforms, so changing from `--i18n` to `--i18n-languages` makes this incorrect for cucumber-ruby and cucumber-jvm.

I think we should simply get rid of the option and refer users to the docs.

aslakhellesoy: See https://github.com/cucumber/cucumber/issues/783

aslakhellesoy:
Should we just close this, or do we want to make the docs language specific? WDYT @mlvandijk ?

mlvandijk:
@aslakhellesoy We don't currently have different language tabs on the Gherkin page(s) as Gherkin should be programming language agnostic.

We could just remove this section imho. - what do you think?

mlvandijk:
Thanks for pointing out the inconsistency though @brazilianbytes!

aslakhellesoy:
There are two suggestions here:

1) Remove the docs about `--i18n`
2) Remove `--i18n` from Cucumbers - see https://github.com/cucumber/cucumber/issues/783

I think we should remove both.

mlvandijk:
@aslakhellesoy Can we point them anywhere to see up to date available languages (rather than duplicate in documentation?). Kind of like we do for DataTables (points to datatable README from the docs)

aslakhellesoy:
Yes @mlvandijk here: https://cucumber.io/docs/gherkin/reference/#gherkin-dialects

mlvandijk:
Perfect! Let's link there. @brazilianbytes - would you be willing to do this change instead?

mlvandijk:
Closed in favor of cucumber/common#431

PR #423, opened 2019-11-27
updates exception path to cucumber.api.PendingException()

EasyDanger:
The example text did not correspond to what the terminal returns, leading to confusion for new people following the tutorial not noticing that they need to copy the snippets from their terminal in order to have access to the given exception, and not from the tutorial.


aslakhellesoy:
It sounds like you are using an old version. What version are you on?

aslakhellesoy:
Closing this, as it's referencing an older API. The docs are referencing the latest API.

Issue #389, opened 2019-10-09
More inclusive language in the docs


carolstran:
:wave: I noticed that there are many instances of the word "simple" (or similar variations) in your documentation. Unfortunately, words like this can make people feel frustrated and isolated while reading (myself included, hence the issue) - especially if they are facing issues. 

If it's alright with your team, I'd like to take a shot at updating the content to make it more inclusive by removing instances of "simple", "easy", etc. 😊 

Also if you're interested or want more information on this topic, I'd highly recommend watching Jim Fisher's [Don't Say Simply](https://www.youtube.com/watch?v=gsT2BBWBVmM) talk from Write the Docs Prague 2018 ✍️ 🤓 

Thank you for the work you do and please let me know if you need more details!

luke-hill:
Definitely agree with this. I actually altered a couple of those recently myself in some docker information. And it's likely I'm guilty of writing it in a few other places.

I'll happily review and merge any PR like this, as it's something I'm quite keen on

carolstran:
Awesome! I'm on it 🚀

mlvandijk:
Yes please @carolstran! I was not aware we were doing that and we obviously intend to help, not frustrate. Multiple PR's would also be welcome for this, if you'd like to take it one page/section at a time. 
If you have any question, please join us on [Slack](https://cucumberbdd-slack-invite.herokuapp.com/) in the #docs channel.

mlvandijk:
Thanks so much for pointing that out!

PR #425, opened 2019-12-20
Update to 10 minute tutorial

### LINKED ISSUE #367: 10-Minute tutorial uses different versions of Cucumber
Looks like the 10 Minute tutorial is using:

`   "-DarchetypeVersion=4.2.6.1"               \`  for maven, `def cucumberVersion = '4.3.1'` for gradle, and imports from: 4.7:

```
import io.cucumber.junit.Cucumber;
import io.cucumber.junit.CucumberOptions;
```

It'd be best to pin the tutorial against 4.2.6.

https://cucumber.io/docs/guides/10-minute-tutorial/

anna-engle:
Removed the snippet for adding the `RunCucumberTest.java` file completely, as version 4.2.6 automatically creates this file for you when you generate the archetype.

Fixes: #367

### DIFF
--- a/content/docs/guides/10-minute-tutorial.md
+++ b/content/docs/guides/10-minute-tutorial.md
@@ -160,23 +160,6 @@ test {
 }
 ```
 
-Then add a file `src/test/java/hellocucumber/RunCucumberTest.java` inside the project
-to enable JUnit 5 integration:
-
-```java
-package hellocucumber;
-
-import io.cucumber.junit.Cucumber;
-import io.cucumber.junit.CucumberOptions;
-import org.junit.runner.RunWith;
-
-@RunWith(Cucumber.class)
-@CucumberOptions(plugin = "pretty", features = "src/test/resources/hellocucumber")
-public class RunCucumberTest
-{
-}
-```
-
 If you have not already, open the project in IntelliJ IDEA:
 
 * **File -> Open... -> (Select build.gradle)**
--- a/content/docs/guides/10-minute-tutorial.md
+++ b/content/docs/guides/10-minute-tutorial.md
@@ -125,7 +125,7 @@ Open the project in IntelliJ IDEA:
 
 **With Gradle**
 
-The easiest way to create this sample Cucumber project using Gradle is to convert the above generated Maven archetype into a Gradle project.
+One way to create this sample Cucumber project using Gradle is to convert the above generated Maven archetype into a Gradle project.
 
 Run the following command from the `hellocucumber` directory:
 
--- a/content/docs/guides/10-minute-tutorial.md
+++ b/content/docs/guides/10-minute-tutorial.md
@@ -125,40 +125,29 @@ Open the project in IntelliJ IDEA:
 
 **With Gradle**
 
-You can use the following `build.gradle` file as an example template. Use your IDE
-to start a new Gradle-based project.
+The easiest way to create this sample Cucumber project using Gradle is to convert the above generated Maven archetype into a Gradle project.
 
+Run the following command from the `hellocucumber` directory:
+
+```shell
+gradle init
+```
+
+Add the following Task to your `build.gradle` file:
 ```groovy
-apply plugin: 'java'
-sourceCompatibility = 11
-targetCompatibility = 11
-
-group = "hellocucumber"
-version = "1.0"
-
-// Versioning of dependencies
-wrapper.gradleVersion = '5.5.1'
-def cucumberVersion = '{{% version "cucumberjvm" %}}'
-def junitVersion = '5.5.0'
-
-repositories {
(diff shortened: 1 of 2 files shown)


mpkorstanje:
I think that part of code was part of the gradle tutorial.

Maybe we should remove the Gralde tutorial parts. The interleaving between Maven and Gradle makes this really messy and hard to follow.

@mlvandijk wdyt?

mpkorstanje:
And I've made a `4.8.0.0`  version of the archetype. Might be useful.

anna-engle:
@mpkorstanje I updated the archetype version to 4.8.0.0 and the Gradle section to include the command to convert from a Maven project.  Instead of using the template provided in the tutorial, we can add the configs/dependencies/task necessary to build.gradle.

content/docs/guides/10-minute-tutorial.md
```diff
@@ -125,57 +125,29 @@ Open the project in IntelliJ IDEA:
...
+        javaexec {
+            main = "io.cucumber.core.cli.Main"
+            classpath = configurations.cucumberRuntime + sourceSets.main.output + sourceSets.test.output
+            args = ['--plugin', 'pretty', '--glue', 'gradle.cucumber', 'src/test/resources']
+        }
+    }
 }
 ```
+Note that you also need to add the necessary dependencies/configurations to `build.gradle` depending on which version of Gradle you are using.
+See the [Build Tools](https://cucumber.io/docs/tools/java/#gradle) section.
```
mpkorstanje: It's not done consistently throughout the documentation but let's use a relative URL here. They're shorter, more readable and guaranteed to stay in the same domain.

```suggestion
See the [Build Tools](/docs/tools/java/#gradle) section.
```

anna-engle:
@mpkorstanje I agree with what you're saying on it being hard to follow, especially now that this tutorial includes js/kotlin/ruby implementations.  Maybe we could implement something like the following: https://www.cockroachlabs.com/docs/stable/training/cluster-startup-and-scaling.html (see the Mac and Linux buttons)

content/docs/guides/10-minute-tutorial.md
```diff
@@ -125,57 +125,29 @@ Open the project in IntelliJ IDEA:
 
 **With Gradle**
 
-You can use the following `build.gradle` file as an example template. Use your IDE
-to start a new Gradle-based project.
+The easiest way to create this sample Cucumber project using Gradle is to convert the above generated Maven archetype into a Gradle project.
```
mpkorstanje: With #389 in mind I think we should rephrase:

```suggestion
One way to create this sample Cucumber project using Gradle is to convert the above generated Maven archetype into a Gradle project.
```

mpkorstanje (approved):
Awesome! This looks much better already. Cheers. If nobody else has any comments I'll merge this in a few days.

I've added a few suggested edits. I think Github should let you commit them automatically but I'm not sure (I've not used this feature much yet).

> Maybe we could implement something like the following: https://www.cockroachlabs.com/docs/stable/training/cluster-startup-and-scaling.html (see the Mac and Linux buttons)

That would be nice. I'm not sure how easy it would be too add. I do know that @mlvandijk and @mattwynne have been fighting with Hugo to make the current language thing possible.

Perhaps we can add Gralde and Maven as language options?

content/docs/guides/10-minute-tutorial.md
annabaker: Making language more inclusive

aslakhellesoy:
Hi @annabaker,

Thanks for your making your first contribution to Cucumber, and welcome to the Cucumber committers team! You can now push directly to this repo and all other repos under the cucumber organization! 🍾

In return for this generous offer we hope you will:

* ✅ Continue to use branches and pull requests. When someone on the core team approves a pull request (yours or someone else's), you're welcome to merge it yourself.
* 💚 Commit to setting a good example by following and upholding our [code of conduct](https://cucumber.io/conduct) in your interactions with other collaborators and users.
* 💬 Join the [community Slack channel](http://cucumber.io/support#slack) to meet the rest of the team and make yourself at home.
* ℹ️ Don't feel obliged to help, just do what you can if you have the time and the energy.
* 🙋 Ask if you need anything. We're looking for feedback about how to make the project more welcoming, so please tell us!

On behalf of the Cucumber core team,
Aslak Hellesøy
Creator of Cucumber

mpkorstanje:
Cheers!

mlvandijk:
@mpkorstanje @annabaker I'd love to have options for Maven/Gradle (and possibly IDE) in the docs, but am not sure how to do so, especially in combination with the current language sections.
I like the way Spring does this, for example: https://spring.io/guides/gs/rest-service/

PR #428, opened 2019-12-24
#314 - adding information on Maven profiles (Fixes #314)

### LINKED ISSUE #314: Clarify use of (Maven) profiles on JVM
Cucumber profiles are not available on JVM:
https://docs.cucumber.io/cucumber/configuration/#profiles

However it is possible to use [Maven profiles](https://maven.apache.org/guides/introduction/introduction-to-profiles.html) to pass configuration options. 

We should at minimum add this to the page, with a link to the Maven page (and possibly a Gradle page, if Gradle offers the same functionality).
It would be even better if we could provide an example of how that would work.

(Source: https://stackoverflow.com/questions/55341945/how-do-i-use-cucumber-profile-to-dynamically-use-different-environment-hostname)

anna-engle:
https://github.com/cucumber/docs.cucumber.io/issues/314
Added to the profiles section for the JVM a bit about how to set something up using Maven profiles.  I couldn't get a working example done in Gradle, but I linked to some documentation I found on mimicking similar behavior.

### DIFF
--- a/content/docs/cucumber/configuration.md
+++ b/content/docs/cucumber/configuration.md
@@ -325,12 +325,8 @@ output.
 {{% /block %}}
 
 ## Default Profile
-{{% block "java" %}}
+{{% block "java,kotlin" %}}
 Cucumber profiles are not available on Cucumber-JVM. See above.
-{{% /block %}}
-
-{{% block "kotlin" %}}
-Cucumber profiles are not available in Kotlin.  See above.
 {{% /block %}}
 
 {{% block "javascript" %}}
--- a/content/docs/cucumber/configuration.md
+++ b/content/docs/cucumber/configuration.md
@@ -231,7 +231,48 @@ This is just a convention though; Cucumber will pick them up from any file{{% te
 # Profiles
 
 {{% block "java" %}}
-Profiles are not available in Java.
+Profiles are not available in Java.  However, it is possible to set configuration options using [Maven profiles](https://maven.apache.org/guides/introduction/introduction-to-profiles.html).
+
+For instance, we can configure separate profiles for scenarios which are to be run in separate environments like so:
+
+``` xml
+    <profiles>
+        <profile>
+          <id>dev</id>
+            <properties>
+                <cucumber.options>--tags @dev</cucumber.options>
+                <base.url>http://dev.base.url.to.application</base.url>
+            </properties>
+        </profile>
+        <profile>
+          <id>qa</id>
+            <properties>
+                <cucumber.options>--tags @qa</cucumber.options>
+                <base.url>http://qa.base.url.to.application</base.url>
+            </properties>
+        </profile>
+    </profiles>
+
+    <build>
+        <plugins>
+            ...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.0.0-M4</version>
+                <configuration>
+                    <systemPropertyVariables>
+                       <cucumber.options>${cucumber.options}</cucumber.options>
+                    </systemPropertyVariables>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+```
+
+To mimick similar behavior using Gradle, see the Gradle docs on [Migrating Maven profiles and properties](https://docs.gradle.org/current/userguide/migrating_from_maven.html#migmvn:profiles_and_properties).
+
+
 {{% /block %}}
 
 {{% block "kotlin" %}}
@@ -293,7 +334,9 @@ output.
 
 ## Default Profile
 {{% block "java" %}}
-Profiles are not available in Java.
+Profiles are not available in Java.  See above.
+
+
 {{% /block %}}
 
 {{% block "kotlin" %}}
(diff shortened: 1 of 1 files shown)


content/docs/cucumber/configuration.md
```diff
@@ -231,7 +231,48 @@ This is just a convention though; Cucumber will pick them up from any file{{% te
...
+
+For instance, we can configure separate profiles for scenarios which are to be run in separate environments like so:
+
+``` xml
+    <profiles>
+        <profile>
+          <id>dev</id>
+            <properties>
+                <cucumber.options>--tags @dev</cucumber.options>
+                <base.url>http://dev.base.url.to.application</base.url>
```
mpkorstanje: Is the base url needed in this example?

content/docs/cucumber/configuration.md
```diff
@@ -231,7 +231,48 @@ This is just a convention though; Cucumber will pick them up from any file{{% te
...
+Profiles are not available in Java.  However, it is possible to set configuration options using [Maven profiles](https://maven.apache.org/guides/introduction/introduction-to-profiles.html).
+
+For instance, we can configure separate profiles for scenarios which are to be run in separate environments like so:
+
+``` xml
+    <profiles>
+        <profile>
+          <id>dev</id>
+            <properties>
+                <cucumber.options>--tags @dev</cucumber.options>
```
mpkorstanje: When I'm using multiple tags this may interact with mavens property interpolation. If I am not mistaken `--tags @dev and not @ignore` will be seen as the property `dev and not ` and maven will try to interpolate it. 

https://maven.apache.org/shared/maven-filtering/

mpkorstanje (commented):
I think I've seen this on Stackoverflow. :laughing:

content/docs/cucumber/configuration.md
mpkorstanje: I've never tried this but it may be possible to avoid the interpolation by using the html entity `&#64;` for `@` instead.

content/docs/cucumber/configuration.md
```diff
@@ -231,7 +231,48 @@ This is just a convention though; Cucumber will pick them up from any file{{% te
 # Profiles
 
 {{% block "java" %}}
-Profiles are not available in Java.
+Profiles are not available in Java.  However, it is possible to set configuration options using [Maven profiles](https://maven.apache.org/guides/introduction/introduction-to-profiles.html).
```
mlvandijk: Should we be extra clear here and say that the "Cucumber" feature of profiles is not available on Cucumber-jvm?

content/docs/cucumber/configuration.md
```diff
@@ -231,7 +231,48 @@ This is just a convention though; Cucumber will pick them up from any file{{% te
 # Profiles
 
 {{% block "java" %}}
```
mlvandijk: Same block will also be relevant for kotlin. Please add `kotlin` here too.

content/docs/cucumber/configuration.md
```diff
@@ -293,7 +334,9 @@ output.
 
 ## Default Profile
 {{% block "java" %}}
-Profiles are not available in Java.
+Profiles are not available in Java.  See above.
```
mlvandijk: Same comment as above regarding "Cucumber" profiles vs Maven profiles as we know them in Java

content/docs/cucumber/configuration.md
mlvandijk: Let's make sure to use an example that will not lead to further questions.

content/docs/cucumber/configuration.md
mlvandijk: It would be if the profile is used to set different environment variables. In which case the tag example above would not be needed. Either one or the other would be sufficient.

mlvandijk (changes requested):

mlvandijk:
Hi @annabaker - thanks for taking the time to add this. I really like it! Could you make some changes to make sure the example is clear and won't lead to additional questions? See @mpkorstanje and my comments.

content/docs/cucumber/configuration.md
anna-engle: I've removed it from the example :)

content/docs/cucumber/configuration.md
anna-engle: I was able to work around this by surrounding the arguments used for --tags in quotes, i.e. --tags "@dev and not @ignore" 

Seemed to work fine when I tested it with Maven.  I modified my example.

content/docs/cucumber/configuration.md
anna-engle: Fixed!

content/docs/cucumber/configuration.md
anna-engle: Fixed :)

content/docs/cucumber/configuration.md
mpkorstanje: Great!

mpkorstanje (approved):
Looks good to me. I've removed the duplicate blocks for Java and Kotlin by combining them in one `{{% block "java,kotlin" %}}`. Please feel free to merge.

anna-engle:
> Looks good to me. I've removed the duplicate blocks for Java and Kotlin by combining them in one `{{% block "java,kotlin" %}}`. Please feel free to merge.

Thanks, @mpkorstanje !

mlvandijk:
Thanks @annabaker!

PR #432, opened 2019-12-29
Update state to be polyglot (Fixes #362)

### LINKED ISSUE #362: The sharing state between steps page mixes languages
The [sharing state between steps](https://cucumber.io/docs/cucumber/state/#sharing-state-between-steps) mixes page different languages and solutions. It is also a topic new users frequently get stuck on. Splitting it out in different languages and making it more tutorial like might help people get unstuck.

An example of a good, if dated, explanation would be: [Sharing state between steps in Cucumber-JVM using PicoContainer](
http://www.thinkcode.se/blog/2017/04/01/sharing-state-between-steps-in-cucumberjvm-using-picocontainer).

mlvandijk:
Updated page to no longer mix programming languages, by added relevant tags and content to each section for each language (where relevant).
Checked page for each programming language. Double check would be appreciated!

### DIFF
--- a/content/docs/cucumber/state.md
+++ b/content/docs/cucumber/state.md
@@ -148,7 +148,7 @@ compile group: 'io.cucumber', name: 'cucumber-picocontainer', version: '{{% vers
 ```
 
 There is no documentation yet, but the code is on [GitHub](https://github.com/cucumber/cucumber-jvm/tree/master/picocontainer).
-For more information, please see [sharing state using Picocontainer](http://www.thinkcode.se/blog/2017/04/01/sharing-state-between-steps-in-cucumberjvm-using-picocontainer).
+For more information, please see [sharing state using PicoContainer](http://www.thinkcode.se/blog/2017/04/01/sharing-state-between-steps-in-cucumberjvm-using-picocontainer).
 {{% /block %}}
 
 {{% block "ruby,javascript" %}} PicoContainer is a Dependency Injection framework for JVM languages. {{% /block %}}
--- a/content/docs/cucumber/state.md
+++ b/content/docs/cucumber/state.md
@@ -1,6 +1,11 @@
 ---
 title: State
 subtitle: Sharing state, isolated state, dependency injection
+polyglot:
+- java
+- javascript
+- ruby
+- kotlin
 ---
 
 It's important to prevent state created by one scenario from leaking into others.
@@ -24,6 +29,7 @@ State can make your steps more tightly coupled and harder to reuse.
 
 ## World object
 
+{{% block "ruby" %}}
 In Ruby, Cucumber runs scenarios in a `World`. By default, the `World` is an instance of `Object`.
 
 All [step definitions](/docs/cucumber/step-definitions) will run in the context of the current `World` instance; a new instance
@@ -91,14 +97,21 @@ World(MyHelper, MyOtherHelpers)
 This will `extend` each new `World` object with those modules.
 
 If you use [Ruby on Rails](/docs/tools/ruby#ruby-on-rails), there is already a `World` set up for you, so you will get
-an instance of `Cucumber::Rails::World`, which is a subclass of `ActionDispatch::IntegrationTest`. This gives you access to a lot of Rails' helper methods.
+an instance of `Cucumber::Rails::World`, which is a subclass of `ActionDispatch::IntegrationTest`. This gives you access
+to a lot of Rails' helper methods.
+{{% /block %}}
 
-Cucumber-js also uses a `World` as an isolated context for each scenario. You can find more information in the
+{{% block "javascript" %}} Cucumber-js uses a `World` as an isolated context for each scenario. You can find more
+information in the
 [cucumber-js documentation on GitHub](https://github.com/cucumber/cucumber-js/blob/master/docs/support_files/world.md).
+{{% /block %}}
+
+{{% block "java,kotlin" %}} JVM languages do not know a "World" object, like Ruby and JavaScript. Instead, you'll need
+to use [Dependency Injection](#dependency-injection).{{% /block %}}
 
 ## Dependency Injection
-If your programming language is Java, you will be writing glue code
-([step definitions](/docs/cucumber/step-definitions) and [hooks](/docs/cucumber/api/#hooks)) in plain old Java classes.
+{{% block "java,kotlin" %}} If your programming language is a JVM language, you will be writing glue code
+([step definitions](/docs/cucumber/step-definitions) and [hooks](/docs/cucumber/api/#hooks)) in classes.
 
 Cucumber will create a new instance of each of your glue code classes before each scenario.
 
@@ -112,10 +125,12 @@ The available dependency injection modules are:
 - [Guice](#guice)
 - [OpenEJB](#openejb)
 - [Weld](#weld)
-- [Needle](#needle)
+- [Needle](#needle) 
+
+{{% /block %}} {{% block "ruby,javascript" %}} Dependency Injection is specific to JVM languages. {{% /block %}}
 
 ### PicoContainer
-
+{{% block "java,kotlin" %}} 
 To use PicoContainer, add the following dependency to your `pom.xml`:
 
 ```xml
@@ -134,9 +149,13 @@ compile group: 'io.cucumber', name: 'cucumber-picocontainer', version: '{{% vers
 
 There is no documentation yet, but the code is on [GitHub](https://github.com/cucumber/cucumber-jvm/tree/master/picocontainer).
 For more information, please see [sharing state using Picocontainer](http://www.thinkcode.se/blog/2017/04/01/sharing-state-between-steps-in-cucumberjvm-using-picocontainer).
+{{% /block %}}
+
+{{% block "ruby,javascript" %}} Picocontainer is a Dependency Injection framework for JVM languages. {{% /block %}}
+
 
 ### Spring
-
+{{% block "java,kotlin" %}} 
 To use Spring, add the following dependency to your `pom.xml`:
 
 ```xml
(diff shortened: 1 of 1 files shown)


content/docs/cucumber/state.md
```diff
@@ -91,14 +97,21 @@ World(MyHelper, MyOtherHelpers)
 This will `extend` each new `World` object with those modules.
 
 If you use [Ruby on Rails](/docs/tools/ruby#ruby-on-rails), there is already a `World` set up for you, so you will get
-an instance of `Cucumber::Rails::World`, which is a subclass of `ActionDispatch::IntegrationTest`. This gives you access to a lot of Rails' helper methods.
+an instance of `Cucumber::Rails::World`, which is a subclass of `ActionDispatch::IntegrationTest`. This gives you access
```
lisacrispin, line 100: I don't understand exactly what changed here. The verbiage looks the same. Maybe just a different line break? But I don't see a line break after "access" on the preview. I know this is just my own ignorance, sorry.

content/docs/cucumber/state.md
```diff
@@ -1,6 +1,11 @@
 ---
 title: State
 subtitle: Sharing state, isolated state, dependency injection
+polyglot:
+- java
+- javascript
+- ruby
+- kotlin
 ---
```
lisacrispin, line 9: Oh, ok, so you made this "polyglot", and that appears to make it have those tabs and have language-specific verbiage on the page appear only for the appropriate language?

content/docs/cucumber/state.md
lisacrispin: I wrote that before I figured out what "polyglot" must mean. So the change is just to make these parts of the page dynamic dependent on the programming language selected, right?

content/docs/cucumber/state.md
```diff
@@ -112,10 +125,12 @@ The available dependency injection modules are:
 - [Guice](#guice)
 - [OpenEJB](#openejb)
 - [Weld](#weld)
-- [Needle](#needle)
+- [Needle](#needle) 
 
-### PicoContainer
+{{% /block %}} {{% block "ruby,javascript" %}} Dependency Injection is specific to JVM languages. {{% /block %}}
```
lisacrispin, line 130: So for JS and Ruby, there is no "dependency injection", so that list of dependency injection modules just has the names and the verbiage about it being specific to JVM languages. It took me a long time to figure this out! I'm sure coders understand when they read this. It just seemed odd to list them out when they aren't even available in those languages.

content/docs/cucumber/state.md
```diff
@@ -134,9 +149,13 @@ compile group: 'io.cucumber', name: 'cucumber-picocontainer', version: '{{% vers
 
 There is no documentation yet, but the code is on [GitHub](https://github.com/cucumber/cucumber-jvm/tree/master/picocontainer).
 For more information, please see [sharing state using Picocontainer](http://www.thinkcode.se/blog/2017/04/01/sharing-state-between-steps-in-cucumberjvm-using-picocontainer).
```
lisacrispin: Should the word 'Picocontainer' be CamelCased as 'PicoContainer', as it is in the doc that is linked to here?

content/docs/cucumber/state.md
```diff
@@ -134,9 +149,13 @@ compile group: 'io.cucumber', name: 'cucumber-picocontainer', version: '{{% vers
...
+{{% /block %}}
 
-### Spring
+{{% block "ruby,javascript" %}} Picocontainer is a Dependency Injection framework for JVM languages. {{% /block %}}
```
lisacrispin: Should it be 'PicoContainer' instead of 'Picocontainer'?

content/docs/cucumber/state.md
```diff
@@ -440,22 +488,27 @@ Feature: Let's write a lot of stuff to the DB
...
-If you're using a [browser automation](/docs/guides/browser-automation) tool that talks to your application over HTTP, the
-transactional approach will not work if your [step definitions](/docs/cucumber/step-definitions) and the web application serving
-HTTP request each have their own database connection.
-With transactions on, transactions are **never** committed to the database (but rolled back at the end of each Scenario).
-Therefore, the web server's connection will never see data from Cucumber, and therefore your browser won't either.
-Likewise, Cucumber's connection won't see data from the web server.
+If you're using a [browser automation](/docs/guides/browser-automation) tool that talks to your application over HTTP,
+the transactional approach will not work if your [step definitions](/docs/cucumber/step-definitions) and the web
+application serving HTTP request each have their own database connection. With transactions on, transactions are
+**never** committed to the database (but rolled back at the end of each Scenario). Therefore, the web server's
```
lisacrispin, line 502: I'm sure this makes more sense to experienced coders. To me, a "transaction" means, I may do one or many database updates, and I don't want them to persist until the whole set of them is done, if anything fails, I want it to roll back. If everything succeeds and the transaction finishes, I want all the updates in it to persist. Or does "transaction" mean something different in this context? 

Is this a true statement: If I am using a browser automation tool like WebDriver, I have to *only* use the database connection from that (WebDriver), I cannot try to do any database updates from Cucumber step definitions, because they will never be persisted? I have no idea how a Selenium WebDriver script gets connected to Cucumber step definitions. Through the automation code?

I would add the word "are" to the phrase in parens - "(but are rolled back at the end of each scenario)"

This seems quite confusing to me, but again, since I have zero experience with this, that's probably normal. Is it worth having an example, or people will just understand this?

content/docs/cucumber/state.md
```diff
@@ -440,22 +488,27 @@ Feature: Let's write a lot of stuff to the DB
...
+connection will never see data from Cucumber, and therefore your browser won't either. Likewise, Cucumber's connection
+won't see data from the web server.
 
 In this case, you will have to turn off database transactions and make sure the data is explicitly deleted before each Scenario.
```
lisacrispin: Make sure *what* data is explicitly deleted?

content/docs/cucumber/state.md
```diff
@@ -476,12 +529,14 @@ With Rails, you can also turn off transaction globally in your `features/support
 ```
 Cucumber::Rails::World.use_transactional_fixtures = false
 ```
+{{% /block %}}
```
lisacrispin, line 532: In the Java,kotlin & JS tabs, there is this statement: "Ruby tools provide specific ways to turn of transactions." - why does it say anything about Ruby in those?

content/docs/cucumber/state.md
```diff
@@ -476,12 +529,14 @@ With Rails, you can also turn off transaction globally in your `features/support
...
 
-## Cleaning Your Database
+{{% block "java,kotlin,javascript" %}} Ruby tools provide specific ways to turn of transactions. {{% /block %}}
```
lisacrispin, line 534: I'm puzzled why there's something about Ruby for the Java, kotlin, JS tabs?

content/docs/cucumber/state.md
mlvandijk: Exactly!

content/docs/cucumber/state.md
mlvandijk: Yes. Polyglot means there are different programming languages on the page (as listed in the "front matter" at the top of the page) and we have dynamic content for each language using the "{{% block "ruby" %}}" and "{{% /block %}}" around a block of text specific for one language (ruby in this example), or a combination of languages (comma-separated).

content/docs/cucumber/state.md
mlvandijk: Yes, exactly. Before the page has information on different languages which was confusing. This PR solves that.

content/docs/cucumber/state.md
mlvandijk: There reason for this is technical; we made it so we cannot have headers inside a language block (although I forgot the reason we did this). Which means that we cannot "hide" the headers that are irrelevant for a particular language, unfortunately. Hence the generic text saying something is not relevant.

content/docs/cucumber/state.md
mlvandijk: It should be spring, actually. Thanks for pointing that out!

content/docs/cucumber/state.md
mlvandijk: Yes, good catch. Thanks!

content/docs/cucumber/state.md
mlvandijk:
```suggestion
{{% block "ruby,javascript" %}} PicoContainer is a Dependency Injection framework for JVM languages. {{% /block %}}
```

content/docs/cucumber/state.md
mlvandijk:
```suggestion
For more information, please see [sharing state using PicoContainer](http://www.thinkcode.se/blog/2017/04/01/sharing-state-between-steps-in-cucumberjvm-using-picocontainer).
```

content/docs/cucumber/state.md
mlvandijk: Yes, thanks!

content/docs/cucumber/state.md
mlvandijk: Yes, "transaction" here is a database transaction. This section describes that if your front end AND your cucumber code use DIFFERENT connections to the database, they might not see each others data if the transaction is rolled back. Does that make sense? I'm wondering whether we could describe this more clearly.

content/docs/cucumber/state.md
mlvandijk:
```suggestion
In this case, you will have to turn off database transactions and make sure the test data is explicitly deleted before each Scenario.
```

content/docs/cucumber/state.md
mlvandijk: Again, because the headers will be visible, but the content will not be relevant for those languages.

content/docs/cucumber/state.md
mlvandijk: See above

mlvandijk:
Hi @lisacrispin, thanks for your review. I made a few small changes. Let me know if I answered your questions sufficiently / whether you have any additional suggestions?

content/docs/cucumber/state.md
lisacrispin: An example would help me, but it is probably unnecessary for the average Cucumber coder.

lisacrispin:
Hi @mlvandijk, it looks good to me, thank you for patiently answering my own questions. There is just one I had without an answer, about what data gets explicitly deleted , referencing line 506. Not a big deal though, that is just for my own edification. :-)

mpkorstanje (commented):
Nice this makes this page allot more readable! Looks good to me.

mpkorstanje (approved):

PR #440, opened 2020-01-05
update docs for Rule, regarding Background

davidjgoss:
When implementing Rule support for JS in https://github.com/cucumber/cucumber-js/pull/1273, noticed that, per [the messages spec](https://github.com/cucumber/cucumber/blob/master/cucumber-messages/messages.proto#L199), a `Rule` can have a `Background`. The docs passively contradict this; looking at the history it has gone back and forth a bit, but latest update seems to be that it _is_ supported.

So this updates the section for `Background` to say it can be used under `Rule`. I've tried to provide some advice about when you might do that, which probably needs tweaking.

Also:

- Makes the warning about incomplete support in Cucumber implementations a more prominent message
- ~~Fixes a minor spacing issue with that message component~~ now in #442

### DIFF
--- a/content/docs/gherkin/reference.md
+++ b/content/docs/gherkin/reference.md
@@ -102,7 +102,7 @@ You can write anything you like, as long as no line starts with a keyword.
 The (optional) `Rule` keyword has been part of Gherkin since v6. 
 
 {{% note "Cucumber Support for Rule"%}}
-Not all Cucumber implementations have added support for the `Rule` keyword.
+Not all Cucumber implementations have finished implementing support for the `Rule` keyword - see [this issue](https://github.com/cucumber/cucumber/issues/869) for the latest status.
 {{% /note %}}
 
 The purpose of the `Rule` keyword is to represent one *business rule* that should be implemented.
--- a/content/docs/gherkin/reference.md
+++ b/content/docs/gherkin/reference.md
@@ -267,7 +267,7 @@ Example: Multiple Givens
 
 ## Background
 
-Occasionally you'll find yourself repeating the same `Given` steps in all of the scenarios in a `Feature` or `Rule`.
+Occasionally you'll find yourself repeating the same `Given` steps in all of the scenarios in a `Feature`.
 
 Since it is repeated in every scenario, this is an indication that those steps
 are not *essential* to describe the scenarios; they are *incidental details*. You can literally move such `Given` steps to the background, by grouping them under a `Background` section.
@@ -276,7 +276,7 @@ A `Background` allows you to add some context to the scenarios that follow it. I
 
 A `Background` is placed before the first `Scenario`/`Example`, at the same level of indentation.
 
-For example, at the `Feature` level:
+For example:
 
 ```gherkin
 Feature: Multiple site support
@@ -305,7 +305,7 @@ Feature: Multiple site support
     Then I should see "Your article was published."
 ```
 
-And, at the `Rule` level:
+`Background` is also supported at the `Rule` level, for example:
 
 ```gherkin
 Feature: Overdue tasks
@@ -328,9 +328,11 @@ Feature: Overdue tasks
   ...
 ```
 
-You can only have one set of `Background` steps per `Feature` or `Rule`. If you need different `Background` steps for different scenarios, consider breaking up your set of scenarios into more `Rule`s or more `Feature`s.
+{{% warn "Use with caution"%}}
+Whilst usage of `Background` within `Rule` is currently supported, it's not recommended, and may be removed in future versions of Gherkin.
+{{% /warn %}}
 
-You would typically place a `Background` at either `Feature` or `Rule` level, but you can do _both_ if you need to; the steps will be run in hierarchical order, so the `Feature` ones first, followed by the `Rule` ones.
+You can only have one set of `Background` steps per `Feature` or `Rule`. If you need different `Background` steps for different scenarios, consider breaking up your set of scenarios into more `Rule`s or more `Feature`s.
 
 For a less explicit alternative to `Background`, check out [conditional hooks](/docs/cucumber/api/#conditional-hooks).
 
--- a/content/docs/gherkin/reference.md
+++ b/content/docs/gherkin/reference.md
@@ -272,12 +272,11 @@ Occasionally you'll find yourself repeating the same `Given` steps in all of the
 Since it is repeated in every scenario, this is an indication that those steps
 are not *essential* to describe the scenarios; they are *incidental details*. You can literally move such `Given` steps to the background, by grouping them under a `Background` section.
 
-A `Background` allows you to add some context to the scenarios that follow it. It can contain one or
-more `Given` steps.
+A `Background` allows you to add some context to the scenarios that follow it. It can contain one or more `Given` steps, which are run before *each* scenario, but after any [Before hooks](/docs/cucumber/api/#hooks).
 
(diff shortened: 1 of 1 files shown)


content/docs/gherkin/reference.md
```diff
@@ -99,7 +99,12 @@ You can write anything you like, as long as no line starts with a keyword.
 
 ## Rule
 
-The (optional) `Rule` keyword has been added in Gherkin v6. (Note that Gherkin 6 has not yet been incorporated into all implementation of Cucumber!)
+The (optional) `Rule` keyword has been part of Gherkin since v6. 
+
+{{% note "Cucumber Support for Rule"%}}
```
mlvandijk, line 104: This is nice; is there any way people can see if their implementation has implemented it yet?

content/docs/gherkin/reference.md
```diff
@@ -262,17 +267,15 @@ Example: Multiple Givens
...
 are not *essential* to describe the scenarios; they are *incidental details*. You can literally move such `Given` steps to the background, by grouping them under a `Background` section.
 
-A `Background` allows you to add some context to the scenarios in the feature. It can contain one or
+A `Background` allows you to add some context to the scenarios that follow it. It can contain one or
 more `Given` steps.
 
-A `Background` is run before *each* scenario, but after any [Before hooks](/docs/cucumber/api/#hooks). In your feature file, put the `Background` before the first `Scenario`.
-
-You can only have one set of `Background` steps per feature. If you need different `Background` steps for different scenarios, you'll need to split them into different feature files.
+A `Background` is run before *each* scenario, but after any [Before hooks](/docs/cucumber/api/#hooks). In your feature file, put the `Background` before the first `Scenario`/`Example`.
```
mlvandijk: How would this work with rule? Can you have multiple rules per feature? If so, can you now have multiple backgrounds in a feature file. Examples would be nice.

content/docs/gherkin/reference.md
mlvandijk, line 104: I like that you added this note.
However, this might lead people to wonder whether thay can / cannot use the Rule keyword in their implementation / version; how would they find out? Can you add generic information on how they could find that out (without having to list them here; which we would have to maintain)

content/docs/gherkin/reference.md
```diff
@@ -303,6 +306,11 @@ Feature: Multiple site support
     Then I should see "Your article was published."
 ```
 
+You can only have one set of `Background` steps at the `Feature` level. If you need different `Background` steps for different scenarios, you have a couple of options:
```
mlvandijk: What about a Background at Rule level? Can we have multiple Rules in one feature file? If so, can we have multiple backgrounds in that file? How would that work (i.e. which will run when and in what order)? Some examples would be nice. They don't have to be here; it could also be a link if we already have examples.

mlvandijk (changes requested):
Thanks for taking the time for this contribution.
Afaik it was decided during the last Cukenfest / Open source day (April 2019) that Rule would *NOT* have a Background, contrary to earlier decisions. If that has since changes, we should update the documentation accordingly. If it hasn't we should update implementations accordingly). It may have changed without me being aware; I just don't know (@aslakhellesoy could you please confirm either way?)

Second, I think the Notes are now harder to read (blue on black), so I'd prefer to revert that. I'd anyway like to keep it out of this PR. If you feel there are front end changes needed (you mentioned also fixing something) please open a separate PR for those changes so we can handle them separately. :)

Thanks!

davidjgoss:
@mlvandijk thanks for the feedback!

> I think the Notes are now harder to read (blue on black)

Yeah that was a mistake, forgot to push a commit - I'll break that out to a separate pull request as you said.

content/docs/gherkin/reference.md
davidjgoss: > Can you have multiple rules per feature?

Yep

> If so, can you now have multiple backgrounds in a feature file

Yes, in effect - but still only one background as a _direct child_ of a feature. I think this does need some rework to communicate that better.

content/docs/gherkin/reference.md
davidjgoss: Good point. In my testing on `cucumber-js`, given this feature:

```gherkin
Feature: a feature
  Background:
    Given some feature-level setup

  Rule:
    Background:
      Given some rule-level setup

    Example:
      When I do a thing
```

The resulting pickle will be:

```gherkin
1. some feature-level setup
2. some rule-level setup
3. I do a thing
```

So before hooks, then feature-level background, then rule-level background, then scenario/example steps. I'm assuming that is the intent...

content/docs/gherkin/reference.md
davidjgoss: You're right, this note raises more questions than answers.

I'm not sure a list would be so bad? What about, for each Cucumber implementation, we show either the version it landed in, or the GitHub issue for it. Then when all are done we can just update to something like "supported in all current versions, you may need to update if you're on an older one"

content/docs/gherkin/reference.md
davidjgoss: Pushed an update that includes:

- An example of usage of `Background` within `Rule`
- Some clearer wording around placement
- An explanation about using both `Feature` and `Rule` level together

content/docs/gherkin/reference.md
davidjgoss: Or could open an issue in the monorepo, track all the detail there, and update the docs when closing it - would that work?

davidjgoss:
FWIW this commit was the last change I could find on this topic https://github.com/cucumber/docs.cucumber.io/commit/85aa891a926196c93b94c05ef98cb6884bf1c425

davidjgoss:
Highlights of a conversation that happened on Slack about this:

> We wanted to disallow Background in Rule, but it requires a change to the Berp parser generator which nobody has had time to do, so it went stale: #590 ...we should tell people it's currently possible, but we strongly discourage using it because a) it will be a mess, and b) it may be removed

davidjgoss:
@mlvandijk I've made changes per your comments:

- The note about `Rule` support in Cucumber now links to a new issue in the monorepo where we can track the status of all official implementations. The idea is once all our implementations have support in their current version, we can update this doc to say that, and close the issue. cc @aslakhellesoy 
- A warning has been added that says using `Background` inside of `Rule` is supported but not recommended, and support might be removed in future. I've kept the example code, but de-emphasised it a bit in relation to the docs around using it at the feature level

content/docs/gherkin/reference.md
```diff
@@ -303,6 +305,35 @@ Feature: Multiple site support
...
+
+    Example: Already used today
+      Given I last used the app earlier today
+      When I use the app
+      Then I am not notified about overdue tasks
+  ...
+```
+
+{{% warn "Use with caution"%}}
+Whilst usage of `Background` within `Rule` is currently supported, it's not recommended, and may be removed in future versions of Gherkin.
```
mlvandijk: Is it supported in all implementations (or JavaScript only)? Maybe point to the issue for removal (even if stale)?

mlvandijk: (Suggestion to make this extra clear; not blocking for merge imho)

davidjgoss: Hmm good point. On the face of it I would say yes, any implementations support `Rule` would currently work with `Background` under it, since it would be a by-product of the compilation to pickles that happens in Gherkin, which should be consistent across implementations.

There are like six things I could have got wrong in that sentence though - maybe one to check with the maintainers for the other languages?

davidjgoss:
@mlvandijk is this good to go now? I think I've addressed everything

mlvandijk (approved):

PR #442, opened 2020-01-07
add css override for message margin

davidjgoss:
## Before

image

## After

image


davidjgoss:
A bit scared about the scale of the diff in the generated CSS - everything looks right so perhaps something to do with indentation or line break setup in Sass?

mattwynne:
Thanks for the conrtribution @davidjgoss! The publishing pipeline has changed a bit recently and I'm not 100% sure if we're still using this CSS now.

@declark can you help?

mlvandijk (approved):
I checked the preview and the margins on the text box / notes are nicer this way. Thanks!

PR #443, opened 2020-01-09
TypeRegistryConfigurer is Deprecated in 5.0.0-RC2 (2019-11-22)

pi-bits:
TypeRegistryConfigurer is Deprecated in 5.0.0-RC2 (2019-11-22) and new way to use custom parameter type is to use
Deprecated
[Core] Deprecate TypeRegistryConfigurer (#1809 Anton Deriabin)
Use @ParameterType and friends instead when using annotation glue.
Use ParameterType and friends instead when using lambda glue.

### DIFF
--- a/content/docs/cucumber/cucumber-expressions.md
+++ b/content/docs/cucumber/cucumber-expressions.md
@@ -80,14 +80,18 @@ output parameters to your own types. Consider this Cucumber Expression:
 If we want the `{color}` output parameter to be converted to a `Color` object,
 we can define a custom parameter type in Cucumber's [configuration](/docs/cucumber/configuration).
 
+{{% block "java8" %}}
+```java
+ ParameterType("color", "red|blue|yellow", Color::new);
+```
+{{% /block %}}
 {{% block "java" %}}
 ```java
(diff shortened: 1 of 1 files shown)


content/docs/cucumber/cucumber-expressions.md
```diff
@@ -80,14 +80,17 @@ output parameters to your own types. Consider this Cucumber Expression:
 If we want the `{color}` output parameter to be converted to a `Color` object,
 we can define a custom parameter type in Cucumber's [configuration](/docs/cucumber/configuration).
 
+{{% block "java8" %}}
+```java
+ ParameterType("color", "red|blue|yellow", Color::new);
```
luke-hill, line 85: from the previous docs, it would be good just to hint the usage of each param? So putting them on multiline with inline comments would help.

mpkorstanje:
I think this should go on top of the other documentation changes related to v5 in https://github.com/cucumber/docs.cucumber.io/pull/398

mpkorstanje:
Thanks for looking to contribute but this was covered by #398 already.

pi-bits:
@mpkorstanje thank you appreciated

PR #453, opened 2020-02-20
Document asterisk for steps in Gherkin reference

davidjgoss:
This has been part of Gherkin for a long time but not really documented. I think I've described it accurately - the example could probably be better though.

cc @mpkorstanje

### DIFF
--- a/content/docs/gherkin/reference.md
+++ b/content/docs/gherkin/reference.md
@@ -260,6 +260,32 @@ Example: Multiple Givens
   But I shouldn't see something else
 ```
 
+### Star Notation
+
+Gherkin also supports a "star" notation for steps, where the `*` character can be used in place of any of the normal step keywords. This can helpful when you have some steps that are effectively a _list of things_, so you can express it more like bullet points where otherwise the natural language of `And` etc might not read so elegantly.
+
+For example:
+
+```gherkin
+Example: List of setup items
+  Given some base context
+  And one thing
+  And another thing
+  And yet another thing
+  When I do something
+```
+
+Could be expressed as:
+
+```gherkin
+Example: List of setup items
+  Given some base context
+  * one thing
+  * another thing
+  * yet another thing
+  When I do something
+```
+
 ## Background
 
 Occasionally you'll find yourself repeating the same `Given` steps in all of the scenarios in a feature.
--- a/content/docs/gherkin/reference.md
+++ b/content/docs/gherkin/reference.md
@@ -260,9 +260,9 @@ Example: Multiple Givens
   But I shouldn't see something else
 ```
 
-### Universal Keyword
+### *
 
-Gherkin also supports a universal keyword for steps, where the `*` character can be used in place of any of the normal step keywords. This can helpful when you have some steps that are effectively a _list of things_, so you can express it more like bullet points where otherwise the natural language of `And` etc might not read so elegantly.
+Gherkin also supports using an asterisk (`*`) in place of any of the normal step keywords. This can be helpful when you have some steps that are effectively a _list of things_, so you can express it more like bullet points where otherwise the natural language of `And` etc might not read so elegantly.
 
 For example:
 
--- a/content/docs/gherkin/reference.md
+++ b/content/docs/gherkin/reference.md
@@ -267,23 +267,25 @@ Gherkin also supports a "star" notation for steps, where the `*` character can b
 For example:
 
 ```gherkin
-Example: List of setup items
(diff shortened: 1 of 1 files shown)


content/docs/gherkin/reference.md
```diff
@@ -260,6 +260,32 @@ Example: Multiple Givens
   But I shouldn't see something else
 ```
 
+### Star Notation
```
mpkorstanje: This could be the star keyword. Not sure though.

davidjgoss: Yeah, I wasn't 100% either. It feels different to the actual keywords, being a sort of placeholder for any of them.

content/docs/gherkin/reference.md
```diff
@@ -260,6 +260,32 @@ Example: Multiple Givens
...
+
+Gherkin also supports a "star" notation for steps, where the `*` character can be used in place of any of the normal step keywords. This can helpful when you have some steps that are effectively a _list of things_, so you can express it more like bullet points where otherwise the natural language of `And` etc might not read so elegantly.
+
+For example:
+
+```gherkin
+Example: List of setup items
```
luke-hill: Lets use an example here of a shopping list (As above uses concretions).

So you could say

```
Given I have eggs
And I have milk
And I have butter
....
```

```
Given I have eggs
* I have butter
* I have milk
```

davidjgoss: Oh yeah, that could work. To expand on it a bit:

```gherkin
Scenario: All done
  Given I am out shopping
  * I have eggs
  * I have milk
  * I have butter
  When I check my list
  Then I don't need anything
```

I feel like you can see clearly how that reads better than `And`s. What do you think?

davidjgoss: Updated

content/docs/gherkin/reference.md
davidjgoss: @aslakhellesoy @sebrose do you have a view on what we call the asterisk thing?

luke-hill: given that all the other words are called keywords and this operates as a keyword, keyword seems a good starting point perhaps?

luke-hill (approved):

content/docs/gherkin/reference.md
davidjgoss: Yeah maybe I'm trying to be too clever in calling it something other than a keyword. In CSS, the `*` is called the "universal" selector which sort of fits I think. So maybe "Universal Keyword"?

marnen: @davidjgoss Is `*` just another way of spelling `And` or `But`? If so, why not just say that?

mpkorstanje: It can also replace `Given`, `When`, `Then`. In fact, Cucumber ignores the keyword completely.

davidjgoss: Yeah given that is a placeholder for _any_ of the keywords then “universal” seems appropriate to me. I’ve updated but will leave this open a bit longer for more feedback

marnen: I’ll admit that I don’t like “universal” in this sense; it seems that it’s trying to be too clever and not clear enough. That’s why I suggested just thinking of it as another way to spell `And`. 

It’s possible, though, that I don’t understand what star notation is meant for, so I may be misunderstanding its significance. (I wasn’t aware that it was in Gherkin before this issue brought it to my attention and, based on the examples given here, I can’t imagine ever using it...)

content/docs/gherkin/reference.md
marnen: I’m not sure I like the doc text here, because to me, `And` reads more naturally than a sort of half-bullet, half-sentence syntax. Generally if I want something to look like bullet points, I’ll put a list in a data table so I don’t need to have this awkward structure...

However, I know we’re not discussing the feature but only its documentation. I just strongly disagree with the idea that the stars are an improvement in the given case. Maybe a better example can be found; I’m just not sure what, because the pressures that Gherkin puts on step syntax are generally opposite to the pressures that would make star steps look good IMHO.

content/docs/gherkin/reference.md
davidjgoss: Okay, so in the interest of getting this in - and addressing the gap in the docs - I've reworded to not refer to it by any name, but rather just call out the asterisk character and what it does.

mpkorstanje:
Cheers!

PR #455, opened 2020-03-03
Improve dependsOn option value in Gradle task

fagia:
Using "testClasses" as value for dependsOn option in Gradle task is more accurate since it will also include test resources (if any)

### DIFF
--- a/content/docs/tools/java.md
+++ b/content/docs/tools/java.md
@@ -96,7 +96,7 @@ Steps:
 
     ```groovy
     task cucumber() {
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
I asked a colleague, they said "this is an improvement"

luke-hill:
a PR superseding this has been merged into `master`. Thankyou for your help in raising this issue though.

Please feel free to contribute elsewhere - Sorry it took so long to review.

PR #462, opened 2020-04-03
Update 10-minute-tutorial.md

zotov:
For solve next error during gradle build:
[Could not get unknown property 'cucumberRuntime' for configuration container of type org.gradle.api.internal.artifacts.configurations.DefaultConfigurationContainer.]
I needed added next  configuration 
```
configurations {
  cucumberRuntime {
    extendsFrom testRuntime
  }
}
```
  to the build.gradle


luke-hill:
Thanks for your contribution. I think this has been fixed in another PR now.

PR #471, opened 2020-05-28
Java Tools: fix system property name for maven

alb-i986:


olleolleolle:
This uses the convention of lowercase with dots instead of camelcase. Looks good!

alb-i986:
I forgot to mention:

https://github.com/cucumber/cucumber-jvm/blob/d14c437294be61bd5c43c72d16d0876b06ec6f96/core/src/main/java/io/cucumber/core/options/Constants.java#L134

alb-i986:
I see now that actually `-Dcucumber.options` was removed recently in #1958.
For now this change is safe, but we need to remember to update the docs again as soon the #1958 is released.

olleolleolle:
Thanks for linking to the commit - and for the vigilance about the next version.

PR #473, opened 2020-06-05
styling: add background to language tabs

### LINKED ISSUE #426: Missing background for language tabs
Some docs pages are polyglot and have versions for Java, JavaScript, Ruby, ect. Each is displayed in its own tab. However the current layout lacks the typical hints that these are in fact tabs. The icons alone visually disappear into the background. 

For example see:

https://cucumber.io/docs/guides/10-minute-tutorial/

image

@declark can you add some hints to make it more obvious these are tabs?

Browser: Firefox 71.0

davidjgoss:
fixes #426 

Using a slightly transparent white seems to work best. Looks like this:

image


olleolleolle (approved):
This change makes a positive visual difference.

mlvandijk (commented):
Definitely an improvement. Thanks @davidjgoss!

Would it be possible to make the tabs actually look like "tabs" though, ie with a line around the top/site, possibly highlighting the one selected? For a quick illustration of what I mean (although not quite): https://freefrontend.com/css-tabs/ 

Btw if that is not something you can do, I'm ok to merge this as is for now as it is an improvement!

davidjgoss:
@mlvandijk no worries I'll give that a go shortly, I think I can make it a bit better.

aslakhellesoy:
This is great! Thanks a lot @davidjgoss

PR #488, opened 2020-06-25
Update links to related tools

enkessler:
Updated the links of a couple tools due to a recent change in repository ownership.


olleolleolle (approved):
Thanks for keeping the links updated <3!

PR #492, opened 2020-07-24
fixes typo in syntax

the-bug:
There was an 'void' missing

### DIFF
--- a/content/docs/cucumber/step-definitions.md
+++ b/content/docs/cucumber/step-definitions.md
@@ -49,7 +49,7 @@ package com.example;
 import io.cucumber.java8.En;
 
 public class StepDefinitions implements En {
-    public StepDefinitions() {
+    public void StepDefinitions() {
         Given("I have {int} cukes in my belly", (Integer cukes) -> {
             System.out.format("Cukes: %n\n", cukes);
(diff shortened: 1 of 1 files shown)


olleolleolle (dismissed):
Change looks good to me!

@mpkorstanje You're Java-fluent, can it be merged, right now?

**Update**: My Java was rusty! Thanks for good review. I'll retract my review.

content/docs/cucumber/step-definitions.md
```diff
@@ -49,7 +49,7 @@ package com.example;
 import io.cucumber.java8.En;
 
 public class StepDefinitions implements En {
-    public StepDefinitions() {
+    public void StepDefinitions() {
```
mpkorstanje, line 52: This is a constructor method. It doesn't need a return type. 

https://docs.oracle.com/javase/tutorial/java/javaOO/constructors.html

mpkorstanje (changes requested):

content/docs/cucumber/step-definitions.md
the-bug: I did not see it... That is an explanation why it did not work...

PR #494, opened 2020-08-04
Update configuration.md

sairam:
attempt to fix https://cucumber.io/docs/cucumber/configuration/ > Ruby . does not render anything

### DIFF
--- a/content/docs/cucumber/configuration.md
+++ b/content/docs/cucumber/configuration.md
@@ -1,7 +1,6 @@
 ---
 title: Cucumber Configuration
 subtitle: How to configure Cucumber
-polyglot: true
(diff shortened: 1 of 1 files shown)


luke-hill:
This does not fix this. However I cannot explain why this does not render.

@jaysonesmith you're pretty good on these sorts of things. I had a look through this page and compared it to another. It makes no sense why the ruby tags wouldn't work.#

cc/ @mlvandijk

mlvandijk:
Since Type Registry is JVM specific I think there is no text on the page for both Ruby and JS. There should be a text for both JS + Ruby saying that Type Registry is for JVM only (like we do on some other pages).

Allthough tbh I don't know whether TypeRegisty is still relevant..

PR #530, opened 2020-10-16
Add missing end of block

PyvesB:
Hello! 👋🏻 

The [configuration sections](https://cucumber.io/docs/cucumber/configuration/) for Ruby and Javascript are empty, and the sections for Java and Scala are incomplete. I believe this is due to the fact that a `{{% block "kotlin" %}}` statement is not properly closed.

#494 can probably be closed when this gets merged in.


olleolleolle (approved):
This looks like a good repair.

PR #551, opened 2020-11-26
Mention support for ``` as doc string delimiter

davidjgoss:
This seems to have been around forever (since at least Gherkin 5) but not documented.

Editor support is poor so have added a note for that.


luke-hill (approved):

luke-hill:
Yes, we also realised this in cucumber-rails. I'll merge as it's definitely something that should be documented.

PR #568, opened 2020-12-18
update package for cucumber Main class

hertzsprung:


aslakhellesoy:
Thanks @hertzsprung !!

PR #575, opened 2021-01-12
Update cucumber-expressions.md

TedyWillian:
Adding an {int} so the number in the expression is compliant with the cucumber(s).
Adding the \ in the matched expressions since the backslash is escaped.


mpkorstanje (approved):

mpkorstanje:
Nicely found! Cheers.

PR #586, opened 2021-02-10
make list of formatters polyglot

### LINKED ISSUE #585: Missing built-in junit reporter or wrong documentation?
The reporting documentation [here](https://cucumber.io/docs/cucumber/reporting/#built-in-reporter-plugins) mentions that there is a built-in junit reporter. But when I use `--format junit`, it doesn't work and complains that module 'junit' is missing.

I tried to look into the error and found that method `getConstructorByType` in [builder.ts](https://github.com/cucumber/cucumber-js/blob/master/src/formatter/builder.ts) doesn' have any mention of `junit`. So, is it a case of wrong documentation?

We have been using a JSON to JUNIT converter previously, which works great for us. But we are now upgrading to cucumber-js 7.0 which warns us that `The built-in JSON formatter is deprecated and will be removed in the next major release`. To be able to use JUNIT converter, the output needs to be in JSON. We don't want to use json-formatter since it has prebuilt executables, which doesn't work for us.

Can someone please guide me on how to get the JUNIT output from cucumber-js v7.x without using the depcrecated JSON format?

Much appreciated!

davidjgoss:
Fixes #585.

Tasks:

- [x] Check if the lists for jvm, ruby and specflow are correct
- [x] Agree what to do about the "Formatters API" section - the examples are ruby only and the event details are I think based on the outdated protocol and should perhaps be replaced by an overview of messages?

Per https://docs.specflow.org/projects/specflow/en/latest/Tools/Reporting.html it seems like SpecFlow has its own reporting model. Should we remove specflow from the language list for this page?

### DIFF
--- a/content/docs/cucumber/reporting.md
+++ b/content/docs/cucumber/reporting.md
@@ -36,12 +36,14 @@ If you don't want to publish your reports to the [Cucumber Reports](https://repo
 generate local reports using one of the following built-in reporter plugins (also known as "formatters"):
 
 {{% text "java,kotlin" %}}
+* `message`
 * `progress`
 * `pretty`
 * `html`
 * `json`
 * `rerun`
 * `junit`
+* `testng`
 {{% /text %}}
 
 {{% text "javascript" %}}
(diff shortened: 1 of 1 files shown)


aslakhellesoy:
Excellent!

> Should we remove specflow from the language list for this page?

Yes I think so.

aslakhellesoy (approved):

PR #603, opened 2021-04-06
[docs] Modify related-tools.md to add cucumber-tables info

madvirus:


mpkorstanje:
Did you know you can use Jackson to transform a row into a POJO using data table types? You may have to implement a few of jackson's mappers for the T+X stuff but overall it will be a better experience.

madvirus:
I did't know that. Thanks. I cancel pull requests.

PR #617, opened 2021-05-18
Update versions.yaml

aurelien-reeves:


aurelien-reeves:
@mpkorstanje should we update junit to 5?

mpkorstanje:
> @mpkorstanje should we update junit to 5?

Not until Cucumber v7 and Junit 5.8 is GA.

aurelien-reeves:
Ok, thanks

So, can we merge?

mpkorstanje:
Sure. 

I haven't been keeping up with the docs project much. But it's not a major change so I expect everything will still be accurate.

PR #632, opened 2021-06-12
add title attr on each language tab link

davidjgoss:


aurelien-reeves:
Rookie question: I do not know that `{{ . }}` syntax. What is it?

luke-hill:
Local level of nesting inside either go or a lot of other templating languages. So it would be equivalent in these diffs to `.Page.Params.polyglot` value during each enumeration. So probably equal to js / java / ruby / kotlin

aurelien-reeves:
> 
> 
> Local level of nesting inside either go or a lot of other templating languages. So it would be equivalent in these diffs to `.Page.Params.polyglot` value during each enumeration. So probably equal to js / java / ruby / kotlin

Thank you.
I understood well what it does, but not from where it comes from.

aurelien-reeves:
@davidjgoss is this still a draft?

davidjgoss:
@aurelien-reeves it improves things although still not perfect - but yeah can be merged

aurelien-reeves (approved):

PR #636, opened 2021-06-22
Document how to define Environment variables

aslakhellesoy:
Lots of people using Cucumber Reports don't know how to define environment variables. Let's document it.


olleolleolle:
A kindness

PR #661, opened 2021-09-07
Clean-up and upgrade some dependencies

aurelien-reeves:
# Description

I noticed in the PR https://github.com/cucumber/docs/pull/648 that we did not need the `mem` and `lodash.mergewith` packages anymore.

So, I removed the `resolution` from `package.json` of cucumber themes, and run a `yarn upgrade` to upgrade `yarn.lock` files.

# Motivation & context

Avoid having code or package that we don't actually need.

## Type of change

- Refactoring/debt (improvement to code design or tooling without changing behaviour)

### DIFF
--- a/themes/cucumber-hugo/package.json
+++ b/themes/cucumber-hugo/package.json
@@ -12,7 +12,6 @@
     "rimraf": "3.0.2"
   },
   "resolutions": {
-    "mem": "^8.0.0",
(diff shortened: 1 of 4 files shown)


olleolleolle (approved):
Good find! The Yarn "resolutions" section needs manual maintenance, and you gave that to us.

PR #663, opened 2021-09-09
Update Hooks documentation

aurelien-reeves:
# Description

Update the cucumber hooks documentation to consider new `BeforeAll`/`AfterAll` hooks, the new `InstallPlugin` ruby hook, and the deprecation of `AfterConfiguration`.

# Motivation & context

There have been changes recently regarding hooks.

Refs. https://github.com/cucumber/cucumber-ruby/pull/1570

## Type of change

- New feature (non-breaking change which adds new behaviour)
- Breaking change (will cause existing functionality to not
  work as expected)

### DIFF
--- a/content/docs/cucumber/api.md
+++ b/content/docs/cucumber/api.md
@@ -230,10 +230,8 @@ Before { scenario: Scenario =>
 const { Before } = require('@cucumber/cucumber')
 ```
 
-Synchronous style:
-
 ```javascript
-Before(function () {
+Before(async function () {
 })
 ```
 
@@ -241,23 +239,6 @@ Before(function () {
 ES6 arrow functions `() => {}` bind `this` to the current context, which prevents
 sharing state between hooks and step definitions.
 {{% /tip %}}
-
-Promise style:
-
-```javascript
-After(function () {
-  return somethingReturningPromise()
-});
-```
-
-Callback style:
-
-```javascript
-After(function (callback) {
-  somethingCallingCallback(callback)
-});
-```
-
 {{% /block %}}
 
 {{% block "ruby" %}}
@@ -370,7 +351,7 @@ After { scenario: Scenario =>
 {{% block "javascript" %}}
 
 ```javascript
-After(function (scenario) {
+After(async function (scenario) {
 })
 ```
 
@@ -471,11 +452,11 @@ BeforeStep { scenario: Scenario =>
 
 {{% text "javascript" %}}
 ```javascript
-BeforeStep(function({pickle, pickleStep, gherkinDocument, testCaseStartedId, testStepId}) {
+BeforeStep(async function({pickle, pickleStep, gherkinDocument, testCaseStartedId, testStepId}) {
     // doSomething
 })
 
-BeforeStep({tags: "@foo"}, function() {
+BeforeStep({tags: "@foo"}, async function() {
     // apply this hook to only specific scenarios
 })
 ```
@@ -528,7 +509,7 @@ AfterStep { scenario: Scenario =>
 
 {{% block "javascript" %}}
 ```javascript
-AfterStep(function({pickle, pickleStep, gherkinDocument, result, testCaseStartedId, testStepId}) {
+AfterStep(async function({pickle, pickleStep, gherkinDocument, result, testCaseStartedId, testStepId}) {
     // doSomething
 })
 ```
@@ -580,7 +561,7 @@ After("@browser and not @headless") { scenario: Scenario =>
 
 {{% block "javascript" %}}
 ```javascript
-Before({tags: '@browser and not @headless'}, function () {
+Before({tags: '@browser and not @headless'}, async function () {
 })
 ```
 {{% /block %}}
@@ -616,16 +597,8 @@ end
 ```javascript
 const { BeforeAll } = require('@cucumber/cucumber');
 
-// Synchronous
-BeforeAll(function () {
+BeforeAll(async function () {
   // perform some shared setup
-});
-
-// Asynchronous Callback
-BeforeAll(function (callback) {
-  // perform some shared setup
-
-  // execute the callback (optionally passing an error when done)
 });
 ```
 {{% /block %}}
@@ -672,16 +645,8 @@ end
 ```javascript
 const { AfterAll } = require('@cucumber/cucumber');
 
-// Synchronous
-AfterAll(function () {
+AfterAll(async function () {
   // perform some shared setup
-});
-
-// Asynchronous Callback
-AfterAll(function (callback) {
-  // perform some shared setup
-
-  // execute the callback (optionally passing an error when done)
 });
 ```
 {{% /block %}}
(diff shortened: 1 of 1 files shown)


aurelien-reeves:
Note: cucumber-ruby with the changes to the hooks has not been released yet, so wait cucumber@7.1.0 before merging this.

content/docs/cucumber/api.md
```diff
@@ -596,68 +596,151 @@ See more documentation on [tags](#tags).
...
 
+  // execute the callback (optionally passing an error when done)
+});
+```
+{{% /block %}}
 
+{{%block "java" %}}
+```java
+@BeforeAll
+public static void beforeAll() {
```
mpkorstanje, line 609: Not yet available in Cucumber JVM v6. Will be in V7.

davidjgoss (commented):
For JS, I would probably just give one example and make it in the `async function` syntax now as that will be the dominant one. WDYT?

content/docs/cucumber/api.md
aurelien-reeves: I though v7 had been released 😅

But actually you have spotted an issue here: the documentation do not care about versions of cucumber. Someone with an older version of cucumber will ever find the appropriate documentation.

mpkorstanje: That's always been like that and usually not a big problem.

mpkorstanje: I did want to release v7. Just haven't gotten around to it yet. :laughing:

aurelien-reeves: So, we'll see at the moment of merging that PR, if v7 not yet released, we'll remove the java part, replacing it with "coming soon"

aurelien-reeves:
> 
> 
> For JS, I would probably just give one example and make it in the `async function` syntax now as that will be the dominant one. WDYT?

I've updated the API documentation to replace all the javascript examples with a single one with an `async function()`. Does it looks good?

content/docs/cucumber/api.md
aurelien-reeves: What about simply redirecting such technical documentation to the repositories like [cucumber-js](https://github.com/cucumber/cucumber-js/blob/main/docs/support_files/hooks.md), and [cucumber-ruby](https://github.com/cucumber/cucumber-ruby/blob/main/features/docs/writing_support_code/hooks/README.md)?

Those remain tech documentations: there is no need to keep those away from the code as opposed to gherkin documentation and references, guides, and other content on cucumber.io/docs

It will be versioned: users will have the possibility to find the appropriate documentation to their versions

It will be easier for us, as developers, to keep it up-to-date and to make it part of our development processes (I mean, like we keep the changelog up-to-date as part of our PR and reviewers take care of this, we could do the same with such tech documentations)

What do you think?

davidjgoss:
@aurelien-reeves yep LGTM on the JavaScript side, thanks!

content/docs/cucumber/api.md
davidjgoss: For my part I think the technical docs on this site should be super thin (closer to concepts than specific APIs) and hand off to relevant docs in the implementation repos. That might just be me though...

aurelien-reeves: yes, I have the same though

content/docs/cucumber/api.md
```diff
@@ -596,68 +577,135 @@ See more documentation on [tags](#tags).
...
+`BeforeAll` run after all scenarios have been executed.
+
+{{%block "ruby" %}}
 ```ruby
-Before do
-  $dunit ||= false  # have to define a variable before we can reference its value
-  return if $dunit  # bail if $dunit is true
-  the_slow_thing    # otherwise do it.
-  $dunit = true     # don't do it again.
+AFterAll do
```
luke-hill: AfterAll

content/docs/cucumber/api.md
```diff
@@ -596,68 +577,135 @@ See more documentation on [tags](#tags).
...
+```scala
+BeforeAll {
+    // doSomething
+}
+```
+{{% /block %}}
 
+### AfterAll
+
+
```
luke-hill: extra line

content/docs/cucumber/api.md
```diff
@@ -596,68 +577,135 @@ See more documentation on [tags](#tags).
...
 
 You can use this hook to extend Cucumber. For example you could affect how features are loaded, or register custom formatters programmatically.
+
+[cucumber-wire](https://github.com/cucumber/cucumber-ruby-wire) is a good example
+of how to use InstallPlugin and what a Cucumber plugin can do.
 {{% /text %}}
 
-{{% text "java,kotlin,scala" %}}Cucumber-JVM does not support `AfterConfiguration` hooks.{{% /text %}}
-{{% text "javascript" %}}Cucumber js does not support `AfterConfiguration` hooks.{{% /text %}}
+{{% text "java,kotlin,scala" %}}Cucumber-JVM does not support `InstallPlugin` hook.{{% /text %}}
```
luke-hill: does not support the `InstallPlugin` hook.

luke-hill (commented):
Minor language tweaks, all seems good though

content/docs/cucumber/api.md
aurelien-reeves: @mpkorstanje should I revert the changes for Java despite the release of cucumber-jvm 7 RC1?

content/docs/cucumber/api.md
```diff
@@ -596,68 +577,134 @@ See more documentation on [tags](#tags).
...
 You can use this hook to extend Cucumber. For example you could affect how features are loaded, or register custom formatters programmatically.
+
+[cucumber-wire](https://github.com/cucumber/cucumber-ruby-wire) is a good example
+of how to use InstallPlugin and what a Cucumber plugin can do.
 {{% /text %}}
 
-{{% text "java,kotlin,scala" %}}Cucumber-JVM does not support `AfterConfiguration` hooks.{{% /text %}}
-{{% text "javascript" %}}Cucumber js does not support `AfterConfiguration` hooks.{{% /text %}}
+{{% text "java,kotlin,scala" %}}Cucumber-JVM does not support the `InstallPlugin` hook.{{% /text %}}
+{{% text "javascript" %}}Cucumber js does not support the `InstallPlugin` hook.{{% /text %}}
```
olleolleolle: Minor: Cucumber.js is the name of the project.

```suggestion
{{% text "javascript" %}}Cucumber.js does not support the `InstallPlugin` hook.{{% /text %}}
```

luke-hill (approved):

davidjgoss (approved):

content/docs/cucumber/api.md
mpkorstanje: Probably best for now. Maybe split them out in to a separate pull request.

aurelien-reeves: 👍 
Will do before merging if cucumber-jvm 7 still not released.

Won't merge this now as cucumber-ruby still need to be released too. But I think we may release it pretty soon.

PR #714, opened 2022-01-08
Reference JUnit 5

mpkorstanje:
Cucumber-JVM can be ran with JUnit 4, TestNG, the CLI and now also JUnit 5.


olleolleolle (approved):
This is very clear now! Thanks for taking care of this.

PR #744, opened 2022-05-03
customeRunnerCliOption.md

akhi9195:
1. When multiple runner classes are present in the project, currently, cucumber cli is running all the runner classes. Instead of this, if cli provides a argument to take the runner class as a argument, that would avoid running all the runner classes. Instead it will run only 1 runner class which is being passed as a argument. 

2. I am using 7.3.1 cucumber dependency version for c-java & c-junit, which is not printing the help commands on the console even after running following 2 commands.
-> java io.cucumber.core.cli.Main --help  [ran from the project location]
-> --help
-> mvn test - Dcucumber.options="--help" [thrown a message build failure, unknown life cycle phase exception]
Hence, please add detailed steps to get the help commands

### 📋 Checklist:


mlvandijk:
Hi @akhi9195 - I'm not sure what you intend to do with this PR.
Did you mean to create an issue instead?

Issue #637, opened 2021-06-23
Language tabs should be URL-aware


mattwynne:
I want to link someone to the Ruby version of https://cucumber.io/docs/cucumber/state/ but the URL doesn't change when I switch language tabs.

I'd like the language tabs to:

1) Switch to the appropriate language tab on opening a page if a querystring parameter is present (e.g. `?lang=ruby`)
2) Update the URL if the language tab is changed (could be dynamic or by just hyperlinking to the new URL)

How would we achieve this?

10xtechie:
@mattwynne, @aurelien-reeves 
I would like to take this on. So if it is still relevant, please assign this to me :)

mattwynne:
Yay! thanks @10xtechie yes it's absolutely still relevant!

https://calendly.com/mattwynne if you want to pair.

mattwynne:
I think the code for the tabs is here: https://github.com/cucumber/docs/blob/main/themes/cucumber-sb/static/js/site.js

10xtechie:
@mattwynne @aurelien-reeves ,
Thanks for assigning this to me! I have created a PR, please take a look.

10xtechie:
> I think the code for the tabs is here: https://github.com/cucumber/docs/blob/main/themes/cucumber-sb/static/js/site.js

@mattwynne ,
You are right, thanks!

PR #765, opened 2022-06-23
issue #637 - changes for URL aware language tabs

10xtechie:
### 🤔 What's changed?

- Switch to the appropriate language tab on opening a page if a query string parameter is provided(e.g. ?lang=java)
- Update the URL if the language tab is changed 

Implementation of an essential feature

### 📋 Checklist:

### DIFF
--- a/themes/cucumber-sb/static/js/site.js
+++ b/themes/cucumber-sb/static/js/site.js
@@ -128,13 +128,19 @@ var supportedLanguages = [
   "scala"
 ]
 
+var defaulLanguage = 'java'
+
 ready(function() {
   var selectedLang = getLangFromUrl();
-  if(selectedLang != '' && selectedLang != null && supportedLanguages.includes(selectedLang)){
-    showOnly(selectedLang)
+  if((selectedLang == '' || selectedLang == null)){
+    if(localStorage.getItem('language') == ''){
+      showOnly(defaulLanguage)
+      localStorage.setItem('language', defaulLanguage)
+    }else{
+      showOnly(localStorage.getItem('language'))
+    }
   }else{
-    //setting default language
-    showOnly("java")
+    showOnly(selectedLang)
   }
 
   each(document, '.tabs li', function(li) {
--- a/themes/cucumber-sb/static/js/site.js
+++ b/themes/cucumber-sb/static/js/site.js
@@ -91,12 +91,56 @@ function showOnly(language) {
   }
 }
 
+function updateQueryParam(selectedLang){
+  let params = location.search.split("&")
+  if(params.length > 1){
+    params.forEach((element, index) => {
+      if(element.includes("lang")){
+        params[index] = 'lang=' + selectedLang
+      }
+    });
+    return location.search = params.join("&")
+  }else{
+    return location.search = 'lang=' + selectedLang
+  }
+}
+
+function getLangFromUrl(){
+  let params = location.search.split("&")
+  var lang  = ''
+  if(params[0].length > 1){
+    params.forEach((element, index) => {
+      if(element.includes("lang")){
+        lang =  params[index].split("=")[1]
+      }
+    });
+  }
+    return lang
+}
+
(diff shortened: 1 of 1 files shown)


themes/cucumber-sb/static/js/site.js
```diff
@@ -91,12 +91,56 @@ function showOnly(language) {
...
+  "scala"
+]
+
 ready(function() {
+  var selectedLang = getLangFromUrl();
+  if(selectedLang != '' && selectedLang != null && supportedLanguages.includes(selectedLang)){
+    showOnly(selectedLang)
+  }else{
+    //setting default language
+    showOnly("java")
```
mattwynne: Might be a good idea to call this default out in a const somewhere.

mattwynne:
This is great, thanks so much!

We could consider storing your preferred language in a cookie, and defaulting to that where the URL doesn't contain a language. I think that might help people navigating around through different pages to stay on the same language. Otherwise, I think a Ruby developer would end up seeing java examples as soon as they switch to a different page.

What do you think? Should we try and tackle that as part of this change, or add it on top in a later iteration?

mattwynne (approved):

10xtechie:
> This is great, thanks so much!
> 
> We could consider storing your preferred language in a cookie, and defaulting to that where the URL doesn't contain a language. I think that might help people navigating around through different pages to stay on the same language. Otherwise, I think a Ruby developer would end up seeing java examples as soon as they switch to a different page.
> 
> What do you think? Should we try and tackle that as part of this change, or add it on top in a later iteration?

@mattwynne ,
Good point, totally agreed. Will update accordingly :)

10xtechie:
@mattwynne,
All changes you suggested have been done! Please take a look :)

themes/cucumber-sb/static/js/site.js
```diff
@@ -91,12 +91,71 @@ function showOnly(language) {
...
+
 ready(function() {
+  var selectedLang = getLangFromUrl();
+  if((selectedLang == '' || selectedLang == null)){
+    if(localStorage.getItem('language') == ''){
+      showOnly(defaulLanguage)
+      localStorage.setItem('language', defaulLanguage)
+    }else{
+      showOnly(localStorage.getItem('language'))
+    }
```
mattwynne, line 141: Maybe these lines (136-141) which, if I'm reading it correctly, are duplicated down on line 146, could be extracted into a single function to improve readability/maintainability?

mattwynne: I want to ship this! Could you consider adding a separate refactoring PR with this nit-pick in it?

mattwynne:
Thanks so much for this!

One thing I notice, on testing it, is that if you move between pages, the URL loses the language, so that if you were to share the link with someone, they might see a different page to you.

I wonder if we should also rewrite the URL when we retrieve the default successfully from local storage, what do you think? Basically treat this like an automatic tab selection.

mattwynne:
Also, I wonder if we could use https://developer.mozilla.org/en-US/docs/Web/API/History/pushState to modify the URL, to avoid a reload of the page?

10xtechie:
@mattwynne ,
I believe rewriting the URL while retrieving the default language successfully from local storage is a good idea, so is using pushState to modify the URL. Can I go ahead and implement them?

aurelien-reeves:
@10xtechie You can go, yes

Please note that I had to make some changes to your code in order to fix an issue: https://github.com/cucumber/docs/pull/771

- detecting available languages is dynamic based on the tabs in the page: because all the pages have not the same available languages. [Some have scala](https://cucumber.io/docs/cucumber/step-definitions/) where [other have not](https://cucumber.io/docs/guides/10-minute-tutorial/), there is even [golang at some places](https://cucumber.io/docs/gherkin/step-organization/), to name a few.
- the default language is not hardcoded anymore
- we make sure that we always activate a language that is part of the available languages

PR #769, opened 2022-06-30
issue #637 - addressing a PR comment for tidying up code

10xtechie:
### 🤔 What's changed?

Tidying up code for better readability as suggested by PR earlier and fixing the issue caused by the return value i.e. "null"; but not null or '', of localStorage.getItem('language').

Making code more maintainable or readable. 

### 📋 Checklist:

### DIFF
--- a/themes/cucumber-sb/static/js/site.js
+++ b/themes/cucumber-sb/static/js/site.js
@@ -91,31 +91,9 @@ function showOnly(language) {
   }
 }
 
-function updateQueryParam(selectedLang){
-  let params = location.search.split("&")
-  if(params.length > 1){
-    params.forEach((element, index) => {
-      if(element.includes("lang")){
-        params[index] = 'lang=' + selectedLang
-      }
-    });
-    return location.search = params.join("&")
-  }else{
-    return location.search = 'lang=' + selectedLang
-  }
-}
-
-function getLangFromUrl(){
-  let params = location.search.split("&")
-  var lang  = ''
-  if(params[0].length > 1){
-    params.forEach((element, index) => {
-      if(element.includes("lang")){
-        lang =  params[index].split("=")[1]
-      }
-    });
-  }
-    return lang
+function updateQueryParam(url, language){
+  url.searchParams.set('lang', language);
+  window.history.pushState({}, '', url);
 }
 
 function showDefaultLang(){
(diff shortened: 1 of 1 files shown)


themes/cucumber-sb/static/js/site.js
```diff
@@ -118,6 +118,14 @@ function getLangFromUrl(){
     return lang
 }
 
+function setDefaultLang(){
+  if(!supportedLanguages.includes(localStorage.getItem('language'))){
+    showOnly(defaulLanguage)
```
luke-hill: I'm aware this was here previously, but is `defaulLanguage` a typo.

10xtechie:
@mattwynne ,
Please take a look at these changes which I made based on your suggestions for making code more readable and maintainable - also fixing a bug.

themes/cucumber-sb/static/js/site.js
10xtechie: @luke-hill ,
Good catch! Undoubtedly it is, thanks! I have just fixed it, please check :)

themes/cucumber-sb/static/js/site.js
```diff
@@ -128,27 +136,17 @@ var supportedLanguages = [
...
 ready(function() {
   var selectedLang = getLangFromUrl();
   if((selectedLang == '' || selectedLang == null)){
-    if(localStorage.getItem('language') == ''){
-      showOnly(defaulLanguage)
-      localStorage.setItem('language', defaulLanguage)
-    }else{
-      showOnly(localStorage.getItem('language'))
-    }
+    setDefaultLang()
```
mattwynne: would `showDefaultLang` be a better name maybe?

themes/cucumber-sb/static/js/site.js
```diff
@@ -128,27 +136,17 @@ var supportedLanguages = [
...
   }else{
     if(supportedLanguages.includes(selectedLang)){
```
mattwynne: Now I think about it, since `null` and `''` are not in the `supportedLanguages` list, maybe we could just merge these two conditionals together? So that we either have a good `selectedLang` and show that one, or we show the "default"

WDYT?

themes/cucumber-sb/static/js/site.js
10xtechie: @mattwynne ,
I think you have a point

themes/cucumber-sb/static/js/site.js
10xtechie: @mattwynne ,
you are right! I have made this change, please take a look at it. Thanks :)

themes/cucumber-sb/static/js/site.js
```diff
@@ -118,6 +118,13 @@ function getLangFromUrl(){
     return lang
 }
 
+function showDefaultLang(){
+  if(!supportedLanguages.includes(localStorage.getItem('language'))){
+    showOnly(defaultLanguage)
+  }else{
+    showOnly(localStorage.getItem('language'))
+  }
```
olleolleolle:
```suggestion
  if(supportedLanguages.includes(localStorage.getItem('language'))){
    showOnly(localStorage.getItem('language'))
  }else{
    showOnly(defaultLanguage)
  }
```

A very minor point: We could flip the if so that the current else case "show only the localStorage-stored language" becomes the primary.

One benefit of that is: We get rid of a `!` and the logic is slightly more straightforward.

10xtechie:
@mattwynne , @aurelien-reeves 
I have addressed your previous concern and suggestion :  https://github.com/cucumber/docs/pull/765#issuecomment-1170810235

mattwynne:
Looks good to me!

PR #816, opened 2022-12-28
Allow raw html in our Markdown templates

### LINKED ISSUE #813: Online documentation is broken and unreadable
### 👓 What did you see?

The online documentation seems to jumble up all of the versions of documentations for all of the languages and many of the sections are repeated multiple times. For example in the Cucumber -> Step Definitions section of documentation you can see following: 

image

In order to check that actual implementation is not buggy I went out and built a the documentation locally and here is the result of the same page:

image

As you can see in the locally deployed website the documentation is rendered perfectly okay. 

In short the problem seems to be that all of the versions of documentation for all languages somehow got mangled together in the deployed website, while locally that is not the case.

### ✅ What did you expect to see?

I expected to see only that documentation which is relevant for the language picked on top

### 🔬 How could we reproduce it?

Just go to the website and open almost any documentation section. The 10-minute tutorial section is particularly bad and includes sections being rendered multiple times.

### 📚 Any additional context?

I was trying to learn Cucumber for my project and this was a real turn-off, been fighting it most of the day until I came to an idea to just grab the source code.

----
 
*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/ISSUE_TEMPLATE/bug_report.md)*

mpkorstanje:

While raw-html sounds like raw-sewage, and you should stay away from it, it is in this case a feature we use heavily. The short codes defined in `themes/cucumber-sb/layouts/shortcodes` decorate our plain text with somewhat useful styling and formatting. This is of course done in HTML.

Fixes: #813 Maybe.

### 📋 Checklist:

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -11,3 +11,6 @@ theme = "cucumber-sb"
 pygmentsUseClassic = true
 pygmentsCodeFences = true
 pygmentsUseClasses = true
+
+[markup.goldmark.renderer]
+  unsafe = true


olleolleolle:
😄 Perfect!

PR #827, opened 2023-01-06
Fix typos & inconsistencies

mlvandijk:
### 🤔 What's changed?

Replace {{text}} with {{block}} in several places (where relevant)
Replace "method or function" with shortcode
Remove unnecessary new lines
Add "Annotated method style" to relevant Java/Kotlin code blocks.
Move {{ /block }} to after Ruby code examples for 2 blocks.
Adding some {{block}}s for "java,kotlin,scala" as needed.
Add white line between start of {{block}} and code block (unless there was text before the code block)
Fix typos
Remove redundant semicolons and {} in Kotlin examples.
Use shortcodes for stepdef-body and expression-parameter
Fix shortcode syntax where needed.

Consistency and looking to fix shortcodes

### 📋 Checklist:

### DIFF
--- a/content/docs/community/contributing.md
+++ b/content/docs/community/contributing.md
@@ -26,6 +26,7 @@ Before you file an [Issue on Github](https://github.com/cucumber/cucumber/issues
 This way you don't waste time by reporting a bug that has already been fixed.
 
 {{% text "ruby" %}}Try the latest released gem:{{% /text %}}
+
 ``` ruby
 gem update cucumber
 ```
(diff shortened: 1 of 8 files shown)


mpkorstanje:
@mlvandijk before you merge this, make sure the short codes don't add extra whitespace, the changes you are making now may work differently once that is fixed.

mlvandijk:
@mpkorstanje the `{{% expression-parameter %}}` does seem to add an extra whitespace, but imho that is still better than multiple new lines when using multiple {{text}} blocks for the different languages. I don;t think I'll be able to fix all the issues in this PR, but it should fix a large portion (incl. returning the missing sections of docs). I can create a follow up issue for the remaining problems for someone else to look at.

mpkorstanje:
It's better, but you're fixing symptoms. It's the short code magic that got broken. You've gotta figure out how the magic used to work and how it works now.

mlvandijk:
Yes, I'm fixing symptoms at least so the text will be shown again. And then I can figure out the magic.

PR #828, opened 2023-01-11
Update who-does-what.md

Sebastan12:
### 🤔 What's changed?
The order of a Feature in the docs. 
This should give it a better flow when reading it
As X i want to do Y in order to achieve Z

inital confusion at the order in which the text was presented.
my brainwas not used to reading "In order to achieve X as person Y i want to do Z"

### 📋 Checklist:

### DIFF
--- a/content/docs/bdd/who-does-what.md
+++ b/content/docs/bdd/who-does-what.md
@@ -62,6 +62,10 @@ number of scenarios.
 
 The `Feature` line names the feature. This should be a short label.
 
+`As a` describes the role of the people/users being served by the feature.
+
+`I want` is a one sentence explanation of what the Feature is expected to do.
+
 `In order to` presents the reason/justification for having the feature. In
 general, this should match to one of the project’s core purposes or “business
 values” such as:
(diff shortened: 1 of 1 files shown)


luke-hill:
This is referring to the connextra above of the sample feature. It's quite common for connextra to use the As a, I want to, So that format. 

However here it's used different preceding labels. But it's still equivalent correct.

If you want to amend the connextra above then that's probably better.

mattwynne:
Thanks for taking the time to suggest a contribution @Sebastan12, but in my view this doesn't make an improvement to the docs.

The format you're proposing is, as Luke says, generally used for expressing a _User Story_ and I generally advise people _against_ having any kind of 1:1 mapping between user stories and features.

There is more context here: https://cucumber.io/blog/bdd/user-stories-and-bdd-features-are-not-stories/

In fact, I personally try to avoid any kind of formulaic structure in the feature description, and encourage people to write free-text prose about why the feature is needed, who it is for, etc.

mattwynne:
More background on the reason the docs are the way they are today: http://antonymarcano.com/blog/2016/08/how-the-industry-broke-the-connextra-template/

Perhaps we should add these references to the docs?

PR #839, opened 2023-02-01
Mark cucumber-ccp as unmaintained

mpkorstanje:
### 🤔 What's changed?

Cucumber-ccp is slowly going unmaintained.


luke-hill:
@mpkorstanje I suggest we just merge this AS IS

PR #850, opened 2023-03-17
Improved event bus performance using UUID generator selectable through SPI

jkronegg:
### 🤔 What's changed?

I added documentation on the UUID generator selection which was added on https://github.com/cucumber/cucumber-jvm/issues/2698 .

Documenting the UUID generator selection method clearly allows the developer to use the `IncrementingUuidGenerator` which  is about 100 times faster than the default `RandomUuidGenerator`. This shorten the feedback loop and reduces the energy consumption.

### 📋 Checklist:

### DIFF
--- a/content/docs/cucumber/state.md
+++ b/content/docs/cucumber/state.md
@@ -475,14 +475,14 @@ Cucumber emits events on an event bus in many cases, e.g.:
 {{% block "java,kotlin" %}}
 Each event has a UUID as unique identifier. The UUID generator can be configured using the `cucumber.uuid-generator` property:
 
-| UUID generator                                        | Features                                | Performance [Millions UUID/second] | Typical usage example                                                                                                                                                                                                                                                          | 
-|-------------------------------------------------------|-----------------------------------------|------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


content/docs/cucumber/state.md
```diff
@@ -460,6 +460,60 @@ The custom object factory can be configured using the `@CucumberOptions` annotat
...
+}
+```
+2. Define the UUID generator as SPI using a file `META-INF/services/io.cucumber.code.eventbus.UuidGenerator` on the classpath (e.g. Maven `resources` directory) containing the generator class name:
+```java
+mypackage.mysubpackage.MyUuidGenerator
+```
+
+This UUID generator will override the default `RandomUuidGenerator`. 
+
+When the `META-INF/services/...` file contains more than one UUID generator or when there are multiple `META-INF/services/...` on the classpath, the `uuid-generator` property must be configured to select the desired generator.
```
mpkorstanje:
```suggestion
When the `META-INF/services/...` file contains more than one UUID generator or when there are multiple `META-INF/services/...` on the classpath, the `cucumber.uuid-generator` property must be configured to select the desired generator.
```

mpkorstanje (approved):
I had a hard time reviewing this. The topic seems a bit out of place for the purpose of the cucumber docs which is to get people to understand the basics. But that is mostly because the plugin system hasn't been explained yet. So this is actually good enough.

PR #873, opened 2023-03-31
Update better-gherkin.md

atdd-bdd:
Added an intermediate style between imperative and declarative

### 🤔 What's changed?

Added a third style between imperative and declarative

It's good to have three perspectives on how to approach an issue.   In many cases, the data in the scenario should be transparent.   The third way allows for transparency, as well as reuse of the scenario in multiple contexts.   

### 📋 Checklist:


content/docs/bdd/better-gherkin.md
```diff
@@ -80,3 +84,39 @@ Scenario: Subscriber with a paid subscription can access both free and paid arti
...
+
+## A third style 
+
+There is a style that is intermediate between these two.   The exact values are specified in the scenario, but the way the user interacts with the system is not.  This style uses stepdef tables with domain terms as the column headers.   The step definitions can  be reused for mulitple scenarios with different data in the table.  In this example, the scenarios for free and paid subscribers use the same step defs.
```
luke-hill: Few minor things to fix:
- Check your formatter here as lots of extra spaces are being added.
- Use `step definitions` instead of stepdef
- `mulitple`* typo

Also these technical items are known as data tables not step definition tables

content/docs/bdd/better-gherkin.md
```diff
@@ -80,3 +84,39 @@ Scenario: Subscriber with a paid subscription can access both free and paid arti
...
+Then articles displayed are 
+| Title           |
+| Free Article 1  | 
+
+Scenario: Subscriber with a paid subscription can access both free and paid articles
+Given articles are: 
+| Title           | For Subscription  |
+| Free Article 1  | Free              |
+| Paid Article 1  | Paid              |
+And user is logged in
```
luke-hill: if you're using a single table, maybe suffix it with `as` i.e. `And the user is logged in as` for both scenarios

content/docs/bdd/better-gherkin.md
```diff
@@ -80,3 +84,39 @@ Scenario: Subscriber with a paid subscription can access both free and paid arti
...
+
+Scenario: Subscriber with a paid subscription can access both free and paid articles
+Given articles are: 
+| Title           | For Subscription  |
+| Free Article 1  | Free              |
+| Paid Article 1  | Paid              |
+And user is logged in 
+| User Name               | Password          | Subscription  |
+| paidPattya@example.com  | validPassword123  | Paid          |
+When articles are displayed
```
luke-hill: This step would then become redundant in both scenarios

luke-hill (commented):
Check your editor as it seems as though some inconsistent spacing is being added.

content/docs/bdd/better-gherkin.md
atdd-bdd: Made the changes as you suggested.    Thanks for the corrections.

content/docs/bdd/better-gherkin.md
atdd-bdd: Made the change

content/docs/bdd/better-gherkin.md
atdd-bdd: I was split on whether the login should become a separate scenario or not.   I was matching the data in the imperative style

There would at least need to be a step that said the logged user had a paid.free subscription.  Keeping that as a data table emphasizes that Subscription is a domain term.    

I could write a section on splitting scenarios, which might come after this one.

luke-hill (commented):
Most of the sentences are good but it's likely you're missing the conjunction "the" in most/all of them (Possibly confusing if you're not a native speaker).

Also if you can remove the redundant steps we can get this merged.

atdd-bdd:
I think this should meet what you asked for.   Let me know if I am still missing something.    Thanks.

luke-hill:
Hi @atdd-bdd Think there's still a couple of things to fix up. Missed your comment sorry

PR #877, opened 2023-04-13
Remove Smart Bear Community recommendation

mpkorstanje:
We use GitHub to track issues and feature requests. When dealing with questions we direct people to the "Get In Touch Page". One of the options includes the Smart Bear Community.

Currently none of the question asked in the Smart Bear Community since the start of 2023 has received any reply. Nor have most of the questions in asked in 2022. 

Unfortunately because the Smart Bear Community does not provide answers people return with repeat questions e.g. https://github.com/cucumber/cucumber-jvm/issues/2723 and then  https://github.com/cucumber/cucumber-jvm/issues/2733.

It would make more sense to direct them to Stackoverflow and Slack exclusively. Questions are still answered there.

### 📋 Checklist:


olleolleolle (approved):
Sensible change!

PR #911, opened 2023-08-12
Doc changes to avoid user confusion of seperating tags with , in expr…

shivam-sehgal:
…ession

### 🤔 What's changed?

I came across https://github.com/cucumber/cucumber-jvm/issues/2776. Although the documentation specifies the expected format for tags, it seems that the pattern used by the user, 'https://github.com/Step1,@step2,' did not generate any error. This discrepancy has the potential to confuse users. To address this, it would be beneficial to add this message to doc

more discussion on this could be found in this [PR](https://github.com/cucumber/tag-expressions/pull/107)

heavily use this library, and wanted to contribute.

If it's related to another issue or bugfix, add it as a reference here, 
e.g. "Fixes cucumber/cucumber-js#99"
--> 
https://github.com/cucumber/cucumber-jvm/issues/2776
https://github.com/cucumber/tag-expressions/pull/107

### 📋 Checklist:

### DIFF
--- a/content/docs/cucumber/api.md
+++ b/content/docs/cucumber/api.md
@@ -969,6 +969,8 @@ A tag expression is an *infix boolean expression*. Below are some examples:
 | `@smoke and @fast`   |                   Scenarios tagged with both `@smoke` and `@fast` |
 | `@gui or @database`  |                Scenarios tagged with either `@gui` or `@database` |
 
+Note: Using a `,` to separate tags in expressions such as `@tag1,@tag2` will cause the entire expression to be parsed as a single tag, potentially leading to unintended outcomes. It's required to separate the tags with space or operators like `and` `or` and `not`.
+
 For even more advanced tag expressions you can use parenthesis for clarity, or
 to change operator precedence:
 


content/docs/cucumber/api.md
```diff
@@ -969,6 +969,8 @@ A tag expression is an *infix boolean expression*. Below are some examples:
 | `@smoke and @fast`   |                   Scenarios tagged with both `@smoke` and `@fast` |
 | `@gui or @database`  |                Scenarios tagged with either `@gui` or `@database` |
 
+Note: Using a `,` to separate tags in expressions such as `@tag1,@tag2` will cause the entire expression to be parsed as a single tag, potentially leading to unintended outcomes. It's required to separate the tags with space or operators like `and` `or` and `not`.
```
luke-hill, line 972: This is legacy and we shouldn't be documenting this. Please use the documentation above.

luke-hill (commented):
Will reject this as it's using legacy information

content/docs/cucumber/api.md
mpkorstanje: @luke-hill I think the purpose was to explain people that there was a different syntax for tag expressions in the past and that they should upgrade theirs. 

@shivam-sehgal this change is missing quite some context to help understand the reader what is going on. Consider a section along the line of "Upgrading from legacy tag expressions" that explains what old tag expressions may have looked like and how they should be rewritten.

luke-hill: In previous areas we've actively removed legacy information, so if we're going down that route (removing legacy info), I'd rather keep consistent than not.

But yes notwithstanding that, if something indicated that this isn't advised then yes I figure that could be ok

PR #938, opened 2023-11-27
Update 10-minute-tutorial.md

rickrhone:
It appears the tutorial was updated to include an example scenario but the documentation was not. This update is to realign the documentation.

### 📋 Checklist:

### DIFF
--- a/content/docs/guides/10-minute-tutorial.md
+++ b/content/docs/guides/10-minute-tutorial.md
@@ -442,11 +442,16 @@ You should see something like the following:
  T E S T S
 -------------------------------------------------------
 
-Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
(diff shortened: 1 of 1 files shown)


content/docs/guides/10-minute-tutorial.md
```diff
@@ -442,11 +442,16 @@ You should see something like the following:
  T E S T S
 -------------------------------------------------------
 
-Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
+Scenario: The example                       # hellocucumber/example.feature:3
```
luke-hill, line 445: This is incorrect because this means you've copied the stuff over. To start with you **should** have an empty project

mpkorstanje: It does look like the tutorial is somewhat out of date though. It still references JUnit 4.

PR #972, opened 2024-03-25
Add GitHub Action to update gherkin-i18n-table

### LINKED ISSUE #936: Automate gherkin-i18n-table based on json of gherkin repo
### 🤔 What's the problem you're trying to solve?

When looking at the docs you can find [a table of the translations](https://cucumber.io/docs/gherkin/languages/), but it's not up to date. It can be updated manually but I think it would be nicer if it is in sync with the gherkin repo itself.

### ✨ What's your proposed solution?

It would be nice if the [gherkin-i18n-table](https://github.com/cucumber/docs/blob/main/layouts/shortcodes/gherkin-i18n-table.html) would be generated based on the [json in the gherkin repo](https://github.com/cucumber/gherkin/blob/main/gherkin-languages.json) which to my understanding is the most up to date version.

### ⛏ Have you considered any alternatives or workarounds?

As mentioned above; updating can be done manually. We can also look at the json instead of the docs.

### 📚 Any additional context?

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/ISSUE_TEMPLATE/feature_request.md)*

WikiRik:
### 🤔 What's changed?

Adds a GitHub Action that check every night for updated to the gherkin-i18n-table and creates a PR when there are updates.
It uses the default GITHUB_TOKEN so no additional needs to be set.

See https://github.com/cucumber/docs/pull/974 for an example of the PR it would create. Because this PR has not been merged yet, it also committed the workflow file. After merging this it would only commit the changes to the HTML table.

Fixes #936

### 📋 Checklist:

### DIFF
--- a/.github/workflows/gherkin-i18n-table.yml
+++ b/.github/workflows/gherkin-i18n-table.yml
@@ -4,9 +4,6 @@ on:
     # https://crontab.guru/every-night-at-midnight
     - cron: "0 0 * * *"
   workflow_dispatch:
-  push: # Temporary so we can test the action
-    branches:
-      - WikiRik/i18n-table-gha
 
 permissions:
   contents: read
--- a/.github/workflows/gherkin-i18n-table.yml
+++ b/.github/workflows/gherkin-i18n-table.yml
@@ -27,4 +27,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
+          commit-message: 'Update gherkin-i18n-table'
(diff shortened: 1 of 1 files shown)


mpkorstanje (approved):
This looks good! Thanks!

mpkorstanje:
And it worked like a charm. Thanks again!

PR #1023, opened 2024-10-01
Update api.md - Describe the combination of 'cucumber.filter.name' and 'cucumber.filter.tags'

s2k:
### What's changed?

Combining cucumber filter options is described in a bit more detail.

### Motivation? 

A colleague and I wondered how the filters are combined, and I think others may have the same question.
Thus, I answered it.


olleolleolle (approved):
This is a nice note!

PR #1038, opened 2024-11-14
CWT-Cucumber: A C++ Cucumber Interpreter

ThoSe1990:
### CWT-Cucumber: A C++ Cucumber Interpreter

This project started as educational project for me but evolved in a good way. It is a native C++20 Cucumber Interpreter. It is in a state where it has few users and therefore I think its a good addition. So feel free to check out the repo. 

### What's your motivation?

I used Cucucmber already in my jobs and it helped me. Unfortunately the C++ versions has some heavy dependencies and when trying to integrate it the first time, its quite an effort to do so. At some point I have worked with interpreters and this seemed for me a good chance to write an interpreter from scratch which was not merely a copy from the literature. So here we are now with CWT-Cucumber. If you wonder why CWT. Its the acronym for coding with thomas which is the name of my blog. I didn't want to copy cucumber.cpp because this name is already in use. 

Edit: Repo link: https://github.com/ThoSe1990/cwt-cucumber

### DIFF
--- a/content/docs/installation/cwt-cucumber.md
+++ b/content/docs/installation/cwt-cucumber.md
@@ -1,0 +1,9 @@
+---
+title: cwt-cucumber
+subtitle: C++
+svg: installation/cplusplus.svg
+implementation: unofficial
+weight: 1313
+---
+
+Please see the [CWT Cucumber on GitHub](https://github.com/ThoSe1990/cwt-cucumber).


luke-hill:
ping @ursfassler - Maybe someone we can get helping with cucumber cpp

